PR #4, opened 2024-07-01
Add man page for klp-build

fgyanz:
Add a new man page explaining how the tool works, the supported commands along with some examples.

The committed roff file should work in any system with `groff` installed along with its corresponding man macro package.

To read and test the documentation, run: `man ./klp-build.1`
To generate a PDF document: `man -Tpdf  ./klp-build.1 > klp-build.1.pdf`


marcosps:
There are some things can we can improve, and/or copy/move to the README.md, but in general it's pretty good! Thanks a lot @fgyanz for creating it for us!

PR #7, opened 2024-07-23
extractor: Add filelocking support for srcdir

lhruska:
This patch adds the required support for sharing the same data folder with multiple instances of klp-build.

The filelock library implements a multiplatform code to create an additional file as a lock to synchronize a folder between multiple processes.

Even though Python does not call "destructor" __del__ in case of an unhandled exception, the library itself takes care of unhandled files, so this location for releasing the lock seems to be enough.

### DIFF
--- a/klpbuild/extractor.py
+++ b/klpbuild/extractor.py
@@ -16,6 +16,7 @@ from collections import OrderedDict
 from pathlib import Path
 from pathlib import PurePath
 from threading import Lock
+from filelock import FileLock
 
 from natsort import natsorted
 
@@ -29,6 +30,9 @@ from klpbuild.templ import TemplateGen
 class Extractor(Config):
     def __init__(self, lp_name, lp_filter, apply_patches, app, avoid_ext, workers=4):
         super().__init__(lp_name, lp_filter)
+
+        self.sdir_lock = FileLock(Path(self.get_data_dir("x86_64"), "sdir.lock"))
+        self.sdir_lock.acquire()
 
         if not self.lp_path.exists():
             raise ValueError(f"{self.lp_path} not created. Run the setup subcommand first")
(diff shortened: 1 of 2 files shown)


marcosps:
I had lots of different ideas about how to approach this, but all of them clash with the goal of this PR: to be able to apply patches while multiple users target the same data-dir. So I believe that this works as it should. One minor nit though: instead of hardcoding "x86_64' we should use ARCH from klpbuild.utils, since I plan to make klp-build be able to be used on multiple archs in the future.

klpbuild/extractor.py
```diff
@@ -30,6 +31,9 @@ class Extractor(Config):
     def __init__(self, lp_name, lp_filter, apply_patches, app, avoid_ext, workers=4):
         super().__init__(lp_name, lp_filter)
 
+        self.sdir_lock = FileLock(Path(self.get_data_dir("x86_64"), "sdir.lock"))
```
marcosps, line 34: it should be using ARCH from klpbuild.utils. Otherwise it's good. Thanks!

marcosps (changes requested):

klpbuild/extractor.py
lhruska: I hardcoded the "x86_64" because we only want to "lock" the src directory because that is the one being modified by the `--apply-patches` argument in the `extractor.py`. By checking all the code executed by using `--apply-patches,` the only src directory being patched is the "x86_64", which is currently also hard-coded. So, adding the same comment should be enough, IMO.

marcosps: I think so. The other solution maybe would be to download the source packages for all archs, now that we don't download the -extra packages (thus the amount of downloads is drastically reduced), but still I don't expect people to use the same data-dir as we do :)

Still thanks a lot for the patch!

marcosps (approved):

PR #8, opened 2024-07-24
Add support for tests files in directories

fgyanz:
Up until now we only supported having a single test file per livepatch, but in some cases a more complex environment is required. For that reason a directory per livepatch will be allowed, containing any files needed to sucessfully test the lp. A `test_script.sh` file will still be used as an entry point at runtime.


marcosps:
One functional change introduced by this PR is that we now require a test_script on kgraft-patches_testcases when formatting patches or do prepare_tests, or the script just raises an exception. I don't mind actually, since we require a test script when testing...

marcosps:
Thanks for the patch!

PR #9, opened 2024-07-25
ksrc.py: Check CVE passed in get-patches

fgyanz:
Check that the input CVE number is valid and
complies with the standard.


marcosps:
Makes sense, thanks!

PR #13, opened 2024-07-30
Do not create a second time the ".config"

fgyanz:
Once the kernel and related binaries get extracted so does the ".config", which gets copied twice to its respective kernel src directory. Only once is good enough.

This commit also fixes a bug introduced in fb40c1c5, where the ".config" file was copied from a source with the same path as the destination, hence leading to a crash.

```
shutil.SameFileError: PosixPath('/home/fgonzalez/src/klp-data/x86_64/usr/src/linux-4.12.14-122.222-obj/x86_64/default/.config') and PosixPath('/home/fgonzalez/src/klp-data/x86_64/usr/src/linux-4.12.14-122.222-obj/x86_64/default/.config') are the same file
```


fgyanz:
Closed. A better [solution](https://github.com/SUSE/klp-build/pull/14) has been merged.

PR #14, opened 2024-07-30
klpbuild: Copy the .config from the right place

marcosps:
When the extraction of rpms is done we should copy the .config from the boot/config-<version>, and then future references of this file should point to linux-obj/.config, per codestreams.


fgyanz:
LGTM! Thanks :+1:

PR #15, opened 2024-08-08
config.py: Support per-user configuration file (draft)

fgyanz:
Add support for a per-user configuration file as the main method of configuring klp-build instead of relying on user-set environment variables.

The new implementation what basically does is:
Upon startup, check if `$HOME/.config/klp-build/config` already exists. If it does, load the configuration. Otherwise, create a default configuration file and same with the files and directories specified in the conf.

The default paths for now are:
work directory -> `$HOME/klp-build/patches`
data directory -> `$HOME/klp-build/data`

### DIFF
--- a/klpbuild/config.py
+++ b/klpbuild/config.py
@@ -87,7 +87,8 @@ class Config:
         config['Paths'] = {'work_dir': workdir,
                            'data_dir': datadir,
                            '## SUSE internal use only ##': None,
-                           '#kernel_src_dir': 'kernel-src/'}
+                           '#kernel_src_dir': 'kernel-src/',
+                           '#ccp_pol_dir': 'kgr-scripts/ccp-pol/'}
 
         logging.info(f"Creating default user configuration: '{self.user_conf_file}'")
         os.makedirs(os.path.dirname(self.user_conf_file), exist_ok=True)
--- a/klpbuild/config.py
+++ b/klpbuild/config.py
@@ -88,7 +88,8 @@ class Config:
                            'data_dir': datadir,
                            '## SUSE internal use only ##': None,
                            '#kernel_src_dir': 'kernel-src/',
-                           '#ccp_pol_dir': 'kgr-scripts/ccp-pol/'}
+                           '#ccp_pol_dir': 'kgr-scripts/ccp-pol/',
+                           '#patches_tests_dir': 'kgraft-patches_testscripts/'}
 
         logging.info(f"Creating default user configuration: '{self.user_conf_file}'")
         os.makedirs(os.path.dirname(self.user_conf_file), exist_ok=True)
@@ -364,9 +365,7 @@ class Config:
         return self.get_data_dir(arch)
 
     def get_tests_path(self):
-        self.kgraft_tests_path = Path(Path().home(), "kgr", "kgraft-patches_testscripts")
-        if not self.kgraft_tests_path.is_dir():
-            raise RuntimeError(f"Couldn't find {self.kgraft_tests_path}")
+        self.kgraft_tests_path = self.get_user_path('patches_tests_dir')
 
         test_sh = Path(self.kgraft_tests_path, f"{self.lp_name}_test_script.sh")
         test_dir_sh = Path(self.kgraft_tests_path, f"{self.lp_name}/test_script.sh")
--- a/klpbuild/config.py
+++ b/klpbuild/config.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2021-2024 SUSE
 # Author: Marcos Paulo de Souza <mpdesouza@suse.com>
 
+import configparser
 import copy
 import gzip
 import io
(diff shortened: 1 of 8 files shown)


marcosps:
Overall I liked the approach. But I think that we could have example of the config file in the commit message, so we could easily understand the end goal of it. Also, I'm adding a comments in the commit directly.

klpbuild/config.py
```diff
@@ -29,13 +30,19 @@
...
-        if not work_dir:
-            raise ValueError("KLP_WORK_DIR should be defined")
+        home = os.getenv("HOME")
+        if not home:
+            raise ValueError("HOME env not defined")
 
-        work = Path(work_dir)
-        if not work.is_dir():
-            raise ValueError("Work dir should be a directory")
+        self.user_conf_file = Path(home, ".config/klp-build/config")
```
marcosps, line 37: I believe that we can expect users to have HOME defined? I mean, the majority of people use the most common distros, and I have never saw anyone without HOME being defined :)

klpbuild/config.py
```diff
@@ -61,20 +68,54 @@ def __init__(self, lp_name, lp_filter, kdir=False, data_dir=None, skips="", work
...
+            self.data = self.get_user_path('data_dir')
 
         # will contain the symbols from the to be livepatched object
         # cached by the codestream : object
         self.obj_symbols = {}
 
         logging.basicConfig(level=logging.INFO, format="%(message)s")
 
+    def setup_user_env(self, basedir):
+        basedir = Path(basedir, "klp-build")
```
marcosps, line 80: At this point we are overriding the argument value? For me it's confusing =/

klpbuild/config.py
```diff
@@ -61,20 +68,54 @@ def __init__(self, lp_name, lp_filter, kdir=False, data_dir=None, skips="", work
...
+        workdir = Path(basedir, "patches")
```
marcosps, line 81: I would prefer having the directory "livepatches" over patches, since it's common to have a "patches" directory. Stating as "livepatches" directory seems more objective.

klpbuild/config.py
```diff
@@ -61,20 +68,54 @@ def __init__(self, lp_name, lp_filter, kdir=False, data_dir=None, skips="", work
...
+        self.user_conf = config
+
+    def get_user_path(self, entry, isdir=True):
+        if entry not in self.user_conf['Paths']:
+            raise ValueError(f"config: '{entry}' entry not found")
+
+        p = Path(self.user_conf['Paths'][entry])
+        if not p.exists():
+            raise ValueError(f"'{p}' file or directory not found")
+        if isdir is True and not p.is_dir():
```
marcosps, line 112: You can check isdir directly, like

```
if isdir:
```

marcosps:
Btw, we should show messages on the screen when we create a new config or load the existing one. I believe that this would inform the user of the tool where they can change these values to point the LP creating or data placement into a different directory.

(Also I'm planning to support creating a LP using /lib/modules and /boot of the host machine, so I would need special values on the data-dir config for it to work properly, so showing the user where to look at the config is useful)

klpbuild/config.py
fgyanz: Actually, it's more like appending. So, initially `basedir` is `$HOME` , and after this line `basedir` becomes `$HOME/klp-build`. I'll try to rewrite this part to avoid misleading people.

klpbuild/config.py
marcosps: It's still the same as the older version. You can use pathlib's Path().home() to get the home, and I think that we don't need to check if it is defined or not...

klpbuild/config.py
```diff
@@ -80,8 +80,10 @@ def setup_user_env(self, basedir):
         datadir = Path(basedir, "data")
 
         config = configparser.ConfigParser(allow_no_value=True)
+
         config['Paths'] = {'work_dir': workdir,
-                           'data_dir': datadir}
+                           'data_dir': datadir,
+                           '#kernel_src_dir': 'kernel-src/ (e.g. https://github.com/SUSE/kernel-source)'}
```
marcosps: I think that kernel-source doesn't contain the CVE branches that we have on kerncvs, so we may remove it. Also, IIRC, current we don't require this value to be defined, so I think that we can relax the check for this setting.

Per my tests it fails if kernel_src_dir is not set, or if it's created by klpbuild but it's commented.

klpbuild/config.py
```diff
@@ -29,13 +30,19 @@
...
-            raise ValueError("KLP_WORK_DIR should be defined")
+        logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+        home = os.getenv("HOME")
+        self.user_conf_file = Path(home, ".config/klp-build/config")
+        if not self.user_conf_file.is_file():
+            logging.warning(f"Warning: user configuration file not found")
+            # If there's no configuration file assume fresh install.
+            # Prepare the system with a default environment and conf.
+            self.setup_user_env(Path(home, "klp-build"))
```
marcosps, line 41: Here I think that we can rename the directory to ``klp``, since klp-build can be confused with the project name. We were using ``kgr`` for historic reasons, but currently we use ``klp`` everywhere.

PR #17, opened 2024-08-16
ksrc: Don't ignore patches without ".patch" suffix

fgyanz:
Currently, when reading the "series.conf" file all lines not ending with ".patch" suffix are being ignored. Many patches, mainly old ones, have no extension in their filename, hence ignored when looking for them. This bug was making klp-build falsely report that no backported patches were found with "get-patches" and "setup" commands (e.g. CVE-2021-47369).

From now on, only commented lines will be ignored in the "series.conf"

For context, issue introduced in: a3e5beb04

### DIFF
--- a/klpbuild/ksrc.py
+++ b/klpbuild/ksrc.py
@@ -259,7 +259,7 @@ class GitHelper(Config):
             #   branch:file/path
             idx = 0
             for patch in patch_files.splitlines():
-                if not patch.endswith(".patch"):
+                if patch.strip().startswith("#"):
                     continue
 
                 idx += 1


marcosps:
@fgyanz thanks for the patch! Can you show me a CVE where this happens?

fgyanz:
Hi @marcosps , sure. As indicated in my first comment, I reproduced the issue with CVE-2021-47369. 

For SLE15.2, SLE15.3 and SLE12.5 the backported patch's filename is:
`patches.suse/s390-qeth-fix-NULL-deref-in-qeth_clear_working_pool_list`

Meanwhile for the rest of the codestreams, the patch's filename is:
`patches.suse/s390-qeth-fix-NULL-deref-in-qeth_clear_working_pool_.patch`

marcosps:
Ok, I think that I understood what you did. What my fix was trying to do was to avoid commits hashes that can appear on series.conf file, IIRC, so maybe we can just filter it our by skipping ``series.conf`` files in this case? Either way, let's merge this change and check the commits that klp-build is able to find for the specific CVE.

If we find any issues with this approach then we may need to adapt it again with a better heuristic, but now for let's merge it. Thanks!

fgyanz:
ah! I didn't know that was possible. I analyzed a few `series.conf` and didn't find such cases, so maybe that's no longer the case? Anyway, good to know, I'll keep and eye on this. Thanks!

Issue #19, opened 2024-09-10
Make klp-build reliably detect is a specific tag/rpm already has the patch that fixes a specific CVE


marcosps:
# What we currently have

This is the workflow that we use to detect if a codestream is vulnerable or not:

- Get the commit hash of the  fix for the current codestream, e.g SLE15-SP6
- Check for all tags that contain that specific commit
  - These codestreams are are not vulnerable
- Check for the CVE number on git-log of a specific codestream (git tag per kernel released)
  - If the CVE is mentioned it means that the specific codestream is not vulnerable, since the patch that fixes it is present on the kernel being checked

But what if the patch that fixes the CVE was backported pro-actively and thus doesn't mention the CVE?

# Other approaches

| Idea | Problem |
|------- | ------------- |
| Check the Git-commit tag on the patch file of the git-tag related to the codestreams | Problem: Couldn't find a way of executing git-grep on a codestream without using ``git-checkout`` to switch to the tag |

Combining the `git-grep` approach would make things more reliable.

fgyanz:
Hi @marcosps,

I created a [branch](https://github.com/fgyanz/klp-build/tree/rework-get-kernel) implementing the proposed idea, in case anyone wants to try it out. I've been testing it so far and seems to work fine, but ofc, that's only assuming we don't encounter a case as the one described in the "Problem" column.

kind regards,
Fernando

marcosps:
@fgyanz please send a PR for your branch so we can test and merge it. But as you mentioned, this doesn't fix the issue that I'm seeing: codestreams that were already released and contain the fix backported, but without the CVE number, meaning that the CVE number was created after the patch was backported and tag created.

Either way we can extend the check later for such cases!

marcosps:
Update the original issue to mention a new idea.

marcosps:
I believe that we currently have in hand everything that we need in order to correct detect if a codestream is vulnerable or not. I was wrong when I thought we werent.

In the case that we have a patch already applied before the CVE entry was created, klp-build would first check which codestream family (SLE12-SP5, SLE15-SP2, ...) has backports, and then check the specific codestreams themselves (12.5u48, 12.5u49, 15.5u10, 15.5u20, ..), so in this case, if a codestream already contains the fix, there won't be any backports done to the codestream families, so we are covered.

With it I believe that we can close this issue. Good work @fgyanz for implementing the last missing piece of the checking code!

PR #20, opened 2024-09-11
ksrc: Improve method of detecting patched kernels

fgyanz:
Currently patched kernels are found by checking which kernel tags contain the patch's commit, previously obtained from the respective SLE main branch.

The new approach directly checks each kernel's tag for the patch itself. By doing so, it covers edge cases in which each kernel contains a different commit. The old method will still be used, but as a fallback.

For more information, see https://github.com/SUSE/klp-build/issues/19

For instance, CVE-2023-52846:

Old way (incorrectly labels `15.6rtu0` as vulnerable):
```
Searching for already patched codestreams...
	15.4u28-29 15.5rtu18 15.5u17 15.6rtu1-2 15.6u2-3
Skipping unaffected codestreams (missing backports):
	12.5u47-59 15.2u39-51 15.3u35-47
All affected codestreams:
	15.4u16-27 15.5rtu3-16 15.5u3-15 15.6rtu0 15.6u0-1
Checking filter and skips...
Checking files, symbols, modules...
Done. Setup finished.
```

New way:
```
Searching for already patched codestreams...

15.4u28 (5.14.21-150400.24.125):
74c7662a961ccba6dac0505059ed7f0dc25e453a

15.4u29 (5.14.21-150400.24.128):
74c7662a961ccba6dac0505059ed7f0dc25e453a

15.5u17 (5.14.21-150500.55.73):
74c7662a961ccba6dac0505059ed7f0dc25e453a

15.5rtu18 (5.14.21-150500.13.64):
74c7662a961ccba6dac0505059ed7f0dc25e453a

15.6u2 (6.4.0-150600.23.14):
cf63988c516f2ec586a9997415dd21876467150b

15.6u3 (6.4.0-150600.23.17):
cf63988c516f2ec586a9997415dd21876467150b

15.6rtu0 (6.4.0-150600.8):
4f322b5c2575152ba2fba232c363ba8d1fee55f7

15.6rtu1 (6.4.0-150600.10.5):
cf63988c516f2ec586a9997415dd21876467150b

15.6rtu2 (6.4.0-150600.10.8):
4f322b5c2575152ba2fba232c363ba8d1fee55f7
cf63988c516f2ec586a9997415dd21876467150b

	15.4u28-29 15.5rtu18 15.5u17 15.6rtu0-2 15.6u2-3
Skipping unaffected codestreams (missing backports):
	12.5u47-59 15.2u39-51 15.3u35-47
All affected codestreams:
	15.4u16-27 15.5rtu3-16 15.5u3-15 15.6u0-1
```

### DIFF
--- a/klpbuild/ksrc.py
+++ b/klpbuild/ksrc.py
@@ -356,7 +356,55 @@ class GitHelper(Config):
 
         return commits
 
-    def get_patched_kernels(self, commits):
+    def get_patched_tags(self, suse_commits):
+        tag_commits = {}
+        patched = []
+        total_commits = len(suse_commits)
+
+        # Grab only the first commit, since they would be put together
+        # in a release either way. The order of the array is backards, the
+        # first entry will be the last patch found.
+        for su in suse_commits:
+            tag_commits[su] = []
+
+            tags = subprocess.check_output(["/usr/bin/git", "-C", self.kern_src,
+                                            "tag", f"--contains={su}",
+                                            "rpm-*"])
+
+            for tag in tags.decode().splitlines():
+                # Remove noise around the kernel version, like
+                # rpm-5.3.18-150200.24.112--sle15-sp2-ltss-updates
+                if "--" in tag:
+                    continue
+
+                tag = tag.replace("rpm-", "")
+                tag_commits.setdefault(tag, [])
+                tag_commits[tag].append(su)
+
+            # "patched branches" are those who contain all commits
+            for tag, b in tag_commits.items():
+                if len(b) == total_commits:
+                    patched.append(tag)
+
+        # remove duplicates
+        return natsorted(list(set(patched)))
+
+    def is_kernel_patched(self, kernel, suse_commits, cve):
+        commits = []
+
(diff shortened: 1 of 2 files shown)


fgyanz:
Thanks, and done. As proposed, I added the old technique as a fallback in case no commits are found with the new method. I also reworked the code, having each technique as separate function.

marcosps:
Hey @fgyanz ! Thanks for the update! I'm checking your new update to the PR, and now we check both the tag and the commit hash. What I had in mind was to first check the commit tag using git grep, and ir we can't find it then to fallback to the old method. I believe that this would make the process faster, since a codestream that matches the check for git log --grep --tags don't need the git tag --contains , and vice versa. If if find in one place, we don't need to check the other.

What do you think about it?

marcosps:
The good news is that with this version it matches the number of affected/patched cases from bsc#1226325 that I tested before, thanks!!!

marcosps:
Well, let's merge it. We can always come back and rework if the delays in searching vulnerable codestreams is taking too much time. Thanks a lot for implementing this functionality!

PR #27, opened 2024-10-15
Fix remaining issues regarding the Codestream class introduction

marcosps:
There were issues regarding the prepare-tests subcommand wrt. mismatch between cs (str()) and cs (Codestream object).


fgyanz:
LGTM! Thanks a lot :)

PR #28, opened 2024-10-16
templ.py: Include linux/livepatch.h header for IBT enabled kernels

marcosps:
Without this header the KLP_RELOC_SYMBOL macro won't be recognized.


fgyanz:
LGTM!

PR #32, opened 2024-10-22
Cleanups all around - and a feature

marcosps:
@fgyanz I've added a new argument to `scan` to search for a --conf, so it's easier for us to spot non supported configurations. Could you take a look to double check the details? I can merge it after you give you your input :)

### DIFF
--- a/klpbuild/config.py
+++ b/klpbuild/config.py
@@ -158,17 +158,7 @@ class Config:
         # Validate only the specified architectures, but check if the codestream
         # is supported on that arch (like RT that is currently supported only on
         # x86_64)
-        for arch in self.conf.get("archs"):
-            if arch not in cs.archs:
-                continue
-
-            kconf = cs.get_boot_file("config", arch)
-            with open(kconf) as f:
-                match = re.search(rf"{conf}=([ym])", f.read())
-                if not match:
-                    raise RuntimeError(f"{name}:{arch} ({cs.kernel}): Config {conf} not enabled")
-
-            conf_entry = match.group(1)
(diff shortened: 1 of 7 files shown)


marcosps:
@fgyanz I solved the issue that you found, it was just checking the archs of the codestream. Just initialized it before checking if the codestream is missing, so this can be used to detect which packages to download.

klpbuild/config.py
```diff
@@ -95,7 +92,6 @@ def setup_user_env(self, basedir):
 
     def load_user_conf(self):
         config = configparser.ConfigParser()
-        logging.info(f"Loading user configuration from '{self.user_conf_file}'")
```
fgyanz, line 99: This is kind of useful, no? I know that due to the current design we get the message several times per run, but still.

klpbuild/ksrc.py
```diff
@@ -444,17 +444,12 @@ def get_patched_kernels(self, codestreams, commits, cve):
                 if not patched and kernel not in suse_tags:
                     continue
 
-                print(f"\n{cs.name()} ({kernel}):")
-
                 # If no patches/commits were found for this kernel, fallback to
                 # the commits in the main SLE branch. In either case, we can
                 # assume that this kernel is already patched.
-                for c in kern_commits if patched else suse_commits:
-                    print(f"{c}")
```
fgyanz, line 452: Do we really need to remove this prints? It helps to track the logic taken by klp-build.

marcosps:
I removed the last patch that reduced the output, so I believe that we improve that in a later stage. Merged!

PR #33, opened 2024-10-24
ibs: Speed-up the bulk download of codestreams

fgyanz:
Use the new "workers" setting to specify the number of threads for downloading/extracting codestreams.


marcosps:
At first the number of workers were bigger, but in my machine I watched slowness when rpm2cpio was running, so I reduced the number of workers. This is specially problematic when we are downloading dozens of codestreams, but yeah, let bump it again and see how it behaves currently.

marcosps (approved):
LGMT!

PR #34, opened 2024-10-24
setup: Remove --codestreams in favor of --filter

marcosps:
--filter exists for a long time, so it's time to retire --codestreams.


fgyanz:
LGTM!

PR #36, opened 2024-10-25
ksrc: Add --quiet to git fetch

marcosps:
Otherwise the command output if quite noisy and adds no value to the procedure.


fgyanz:
LGTM!

PR #40, opened 2024-10-31
Move get_cs_branch to utils, plus some cleanups

marcosps:
These are some cleanups that I had on my local branch for a while, plus solving an issue about recursive import.


fgyanz:
LGTM!

PR #41, opened 2024-11-04
More core to codestreams

marcosps:
More clenaups, move validate_config to Codestream class, and more cleanups. Config is being cleanup, so in the future it will hold only code to read codestreams.json.


fgyanz:
LGTM!

PR #42, opened 2024-11-04
Drop conf.json and more cleanups

marcosps:
Now we moved almost all boilerplate code to the classes the inhertit Config. For now only lp_path is there, but that will change soon.


fgyanz:
LGTM! There are some changes I haven't manually tested yet, but the logic is sound. Thanks for the work and apologies for the delay :)

PR #43, opened 2024-11-25
No config - move even more things out of Config class

marcosps:
With this PR we are mostly done: a ton of things now are moved into util and Codestream classes

### DIFF
--- a/klpbuild/ksrc.py
+++ b/klpbuild/ksrc.py
@@ -12,7 +12,6 @@ from datetime import datetime
 from pathlib import Path
 from pathlib import PurePath
 
-import git
 import requests
 from natsort import natsorted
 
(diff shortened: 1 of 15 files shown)


klpbuild/ksrc.py
```diff
@@ -143,8 +143,8 @@ def get_commits(self, cve):
         print("Fetching changes from all supported branches...")
 
         # Mount the command to fetch all branches for supported codestreams
-        subprocess.check_output(["/usr/bin/git", "-C", self.kern_src, "fetch",
-                                 "--quiet", "--tags", "origin"] +
+        subprocess.check_output(["/usr/bin/git", "-C", str(self.kern_src), "fetch",
+                                 "--quiet", "--tags", "--force", "origin"] +
```
fgyanz, line 148: Can you clarify why do we need `force` here? Is there any advantage to the current behavior? I'm wondering if this will have a negative effect on performance and when fetching from multiple instances of klp-build at the same time.

marcosps: Sure, there was one error when fetching changes from kernel-sources related to so stale/removed tags, IIRC, but as we don't care much about those I just added force so it just bypasses it, instead of errorring out.

fgyanz: ah I see! Alright, makes sense :)

fgyanz:
LGTM! Thanks for all the work!

PR #44, opened 2024-11-27
ksrc: Don't download codestreams data with "scan"

fgyanz:
When running the "`scan`" subcommand, don't download the codestreams data unless the "`--conf`" option is set. This small optimization makes it possible to automate "`klp-build scan`" and integrate it into other tools.

### DIFF
--- a/klpbuild/ksrc.py
+++ b/klpbuild/ksrc.py
@@ -482,7 +482,7 @@ class GitHelper(Config):
 
             cs.set_archs(archs)
 
-            if not cs.get_boot_file("config").exists():
+            if conf and not cs.get_boot_file("config").exists():
                 data_missing.append(cs)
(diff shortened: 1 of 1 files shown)


marcosps:
@fgyanz so in this case we should only check for the codestreams that we have downloaded previously, right? What means that either way we should download them before running scan either way?

I'm asking because I thought that, for scanning, we should scan for all currently available codestreams, and thinking that some of them might be added/removed a day before running scan, he might give wrong output in this case: for recently out-of-support codestreams can be tagged as vulnerable, and by newly added codestreams that might be indeed vulnerable would not be listed by the scan subcommand.

marcosps:
That's true indeed. Thanks!

marcosps (approved):
LGMT!

PR #45, opened 2024-11-29
Minor improvements related to the `scan` subcommand

fgyanz:
Add a protection against git's data corruption when multiple klp-build instances are fetching `kernel-sources`. Note that this doesn't fully shield against race-conditions.

Add definition of the `scan` subcommand.


marcosps (approved):
Just checked the --atomic option, and it's ok to add it. LGTM, thanks!

marcosps (approved):

PR #46, opened 2024-12-02
extractor: Check if make invocation fails

marcosps:
This can happen when the extraction process is interrupted, resulting in missing kernel files.


fgyanz:
LGTM! Thank you for the hard work and well done finding the bug! :)

PR #52, opened 2024-12-09
Pylint: Solve reports on Setup class and on test files

marcosps:
Rework the Setup class to better align with what pylint reported.


fgyanz:
LGTM! Thanks :)

PR #53, opened 2024-12-10
setup: Do not reassign archs with archs.sort()

marcosps:
We don't need to assign archs back aftert sorting, since sort() returns None.


fgyanz:
LGTM! Thanks! :)

PR #54, opened 2024-12-16
ibs: push: Get the codestream branch and rebase it on top of base branch

marcosps:
Until now it was required to the developer to always base the branch on top of one of branches assigned to the codestreams. As the number of base branches increase, it's easier if klp-build could make it automatically.

From now on, it's expected that the bscXXXXXX branch is based on the kgraft-patches' master branch, which doesn't include any code, since klp-build will clone that branch, and rebase it on top of a base branch related to the codestream affected.

For example, for 12.5 codestreams, it will get the branch for bscXXXXX_12.5uXXX (which should contain only the directory for the bscXXX) and rebase on top of origin/master-livepatch-sle12.

Hopefully this change will make the LP developers to check less details when creating a livepatch.

### DIFF
--- a/klpbuild/ibs.py
+++ b/klpbuild/ibs.py
@@ -24,7 +24,7 @@ from natsort import natsorted
 from osctiny import Osc
 
 from klpbuild.config import Config
-from klpbuild.utils import ARCH, ARCHS, get_all_symbols_from_object, get_elf_object, get_elf_modinfo_entry, get_cs_branch
+from klpbuild.utils import ARCH, ARCHS, get_all_symbols_from_object, get_elf_object, get_elf_modinfo_entry, get_cs_branch, get_kgraft_branch
 from klpbuild.utils import filter_cs
 
 class IBS(Config):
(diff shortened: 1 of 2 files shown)


fgyanz:
Great idea, @marcosps ! This will improve a workflow a lot. Just one question. Won't this change break the compatibility with `format-patches` and other commands, if any, that rely on the branchs naming? Up until know we followed the convention of grouping all codestreams in the same branch if they we related. E.g: `bscxxxx_15.6u0_15rtu0 `.

marcosps:
Up until now we relied on the branch names from kgraft-patches repo, and that didn't change. The format-patches also uses it to create the patches, so I don't see why it shouldn't work.

We would still have the branches grouped by codestreams, and that's ok, and changes is the process of creating the branches on kgraft-patches. Until now, besides the grouping, we should still divide the number of branches manually. Now, with this approach, we can rely only on grouping to give us the number of different versions of the LP.

(I believe that I still need to add a new patch in order to get the grouping to consider the template generated source code, instead of relying on the output of klp-ccp without the templates applied).

fgyanz:
ok, I see. There was a misunderstanding on my part. LGTM!

marcosps:
@fgyanz I rebased the change to devel.

PR #55, opened 2024-12-16
ksrc: Skip SLE Micro for now

marcosps:
We still need to adjust our tools around it.


marcosps:
@fgyanz can you please merge this asap? supported.csv will contain MICRO versions soon, and it would break klp-build for now.

PR #56, opened 2024-12-17
Check rpm state

marcosps:
@fgyanz take a look!

### DIFF
--- a/klpbuild/ibs.py
+++ b/klpbuild/ibs.py
@@ -113,18 +113,30 @@ class IBS(Config):
 
         rpm_file = Path(dest, rpm)
         cmd = f"rpm2cpio {rpm_file} | cpio --quiet -uidm"
-        subprocess.check_output(cmd, shell=True, cwd=path_dest)
+        subprocess.check_output(cmd, shell=True, stderr=None, cwd=path_dest)
 
         logging.info(f"({i}/{self.total}) extracted {cs.name()} {rpm}: ok")
 
     def download_and_extract(self, args):
         i, cs, _, _, arch, _, rpm, dest = args
 
-        self.download_binary_rpms(args)
+        # Try to download and extract at least twice if any problems arise
+        tries = 2
+        while tries > 0:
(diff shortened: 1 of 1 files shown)


fgyanz:
Hi @marcosps! Thanks for the patch. I'd like to propose a different way. Instead of catching exceptions during extraction, I think it would be better to just check the hash of the downloaded file. If the hash differ from the one in IBS, we re-download. If the extraction fails, then at least we will be sure it's not due to a corrupted rpm.

marcosps:
About the devel, I though that once I pushed a commit based on devel it would automatically target devel. My bad! About checking the hash, it would involved calling one more subprocess, just to check it. And also getting the hash from the file remotely (that would involve downloading more files), when in the end we only want to have the files for the codestream.

So yes, it seems that your idea seems more ideal, but IMHO, it increases the complexity when the outcome would be the same. For some situations where you have more than one RPM begin downloaded, and it's interrupted, it takes a lot of work to check all of them that are affected, removing the files, starting `setup` again and picking the next one...

So IMHO, this is a big `quality of life` improvement in this case. Maybe we can merge this one and then improving it in the future? What do you think?

klpbuild/ibs.py
```diff
@@ -120,11 +120,23 @@ def extract_rpms(self, args):
...
+                # Remove the downloaded RPMs and try again
+                tries = tries - 1
+                logging.info(f"Problem to extract {rpm}. Downloading it again")
+                Path(dest, rpm).unlink()
 
-        # Do not extract kernel-macros rpm
-        if "kernel-macros" not in rpm:
-            self.extract_rpms((i, cs, arch, rpm, dest))
+        if tries == 0:
+            raise RuntimeError(f"Failed to download {rpm}. Aborting")
```
fgyanz, line 139: s/download/extract/ no? To keep consistency with the prev log. Also, I think it could be useful to print here the `stderr` from the rpm2cpio process. What do you think?

fgyanz:
LGTM!

PR #57, opened 2024-12-23
extractor: Replace '-' for "-" on module name of externalized syms

marcosps:
Otherwise the kallsyms machinary fails to find the symbol.

@fgyanz I've tagged this to main, since it's a fix. I believe that after merged this into main we should also merge into devel.


fgyanz:
LGTM!

PR #58, opened 2024-12-27
Run kgr improvements

marcosps:
General improvements when running the tests. With these changes I believe that we are more flexible to store kgr-test and vms directories in our test servers.


fgyanz:
LGTM!

PR #60, opened 2025-01-03
setup: Always set archs and cve to cs_data

marcosps:
Also, fix the problematic test that was checking a condition that was only being triggered if only one arch was set, but was using all archs wrongly.


fgyanz:
LGTM! Tested and works, thanks :)

PR #61, opened 2025-01-07
Micro support

marcosps:

PR #62, opened 2025-01-08
Reorganize codebase structure

vmezzela:
This Pull Request does not change how things work in `klp-build`, instead it aims at reorganizing the code.

Two new packages have been introduced: `klplib` and `plugins`. The majority of modules have been relocated based on the following reasoning:
- Modules that implement a command (e.g., `setup`, `extract`, `inline`) are placed in the `plugins` package.
- Internal modules and shared utilities are grouped under the klplib package.

However, this reorganization is not complete, as it would require to further split the functionalities of each class. At the moment, I believe that the codebase is not yet ready for such a change. I’d like to first refactor the classes to remove inheritance from Config. Once that’s done, I’ll be able to continue with the reorganization more effectively.


vmezzela:
Rebased on top of #63

vmezzela:
Rebased on top of #61

marcosps (approved):
I've tested locally and it worked as expected. Also, the code is much better organized, thanks a lot for doing it!

fgyanz:
Same here, tested and works as expected. Thanks for the rework! :)

fgyanz (approved):

vmezzela:
Hi, thanks a lot :)

PR #63, opened 2025-01-08
Update supported kernel branches

vmezzela:


marcosps:
Thanks! Checked here and it fixes the test issue.

PR #65, opened 2025-01-21
templ: Create Kbuild.inc file only after generating the LP files

marcosps:
Otherwise it generates an empty Kbuild.inc file.


fgyanz:
LGTM! Tested and works as expected. Thanks!

fgyanz (approved):

PR #66, opened 2025-01-21
ksrc: Filter out commits related to the CVE

fgyanz:
In some edge-cases klp-build was reporting wrong commits as the ones introducing the backported patches for the CVE. Improve the filtering heuristic to catch them.


marcosps (approved):
LGTM! I though about adding a new test for it, but given the nature of testing specific CVEs with specific codestreams is not future proof. Either way we'll catch problematic cases  as soon as we work on new CVEs.

PR #67, opened 2025-01-23
Remove mod dep check

marcosps:


klpbuild/klplib/ibs.py
```diff
@@ -277,13 +277,6 @@ def validate_livepatch_module(self, cs, arch, rpm_dir, rpm):
         lp_mod_path = sorted(rpm_dir.glob("**/*.ko"))[0]
         elffile = get_elf_object(lp_mod_path)
 
-        # Check depends field
-        # At this point we found that our livepatch module depends on
-        # exported functions from other modules. List the modules here.
-        deps = get_elf_modinfo_entry(elffile, "depends")
```
vmezzela, line 283: Hi, sorry for the late review. Just a nitpick. Here we can stop importing `get_elf_modinfo_entry`, it's not used elsewhere in this file.

I can drop it in the next PR.

PR #69, opened 2025-01-23
Drop classes inheritance from Config Class

vmezzela:
This Pull Request removes the dependency that most classes had on the Config class.

Before making changes to `Config`, it was necessary to reorganize how we handle codestreams, specifically the methods used for loading and flushing the codestream file.

## Previously:
- The codestream file was loaded whenever an object of any class was initialized. This led to unnecessary multiple loads due to almost all classes depending on `Config`.
- The codestream file was flushed using the `flush_cs_file` function in `klplib/config` when needed.

## Changes in this PR:
- The load and store logic of the codestream file has been decoupled from `Config`. `klplib/config`  is now responsible only for reading the configuration file and making its values accessible throughout the codebase. The load/store logic has been relocated to `plugin/codestreams`. While this placement is a temporary solution, it allows us to untie the dependency from `Config` and makes the codebase easier to restructure moving forward;
- The dependency that most classes had on the Config class has been removed: now instead of accessing configuration values through `self.` attributes, we can call `get_user_path` or `get_user_settings` from everywhere. We do not have to care if the configuration file has been loaded previously because a decorator will check for us;
- The function `get_tests_path` which was available in `Config` has been moved in `klplib/utils` since it's not strictly related to `Config`;
- Other attributes that could be retrieved through `self.` beforehand, are now available through specific functions in `klplib/utils`:
    - `self.lp_path`  -> `get_workdir(lp_name)`
    - `self.data`       -> `get_datadir()`

## Next Steps
The current placement in `plugin/codestreams` is not ideal, as the plugin directory should focus solely on handling commands passed via `klp-build <command>`. A better, more permanent location for the codestream handling logic should be determined in future iterations.

While we already have the `klplib/codestream` module for the `Codestream` class, moving the logic there might make it less organized. I’m unsure about the best structure here and would appreciate any suggestions or feedback on where this logic should ideally reside to keep the codebase clean and maintainable.

If you prefer splitting this PR I can drop some commits from here to ease the review. :)


klpbuild/plugins/codestreams.py
```diff
@@ -0,0 +1,153 @@
...
+    Get the path to the codestreams JSON file for the specified live patch.
+
+    Args:
+        lp_name (str): The name of the live patch.
+
+    Returns:
+        Path: The path to the codestreams JSON file.
+    """
+    workdir = Config(lp_name).get_user_path('work_dir')
+    return workdir/lp_name/'codestreams.json'
```
marcosps, line 93: Interesting, this works as expected? I mean, up until this point I've being using pathlib to create paths like this, and I never saw this syntax before. I'm ok if this works, but I would like to know if this is >python3.11 or not, so we can start using this syntax from now on.

klpbuild/plugins/codestreams.py
```diff
@@ -0,0 +1,153 @@
...
+def set_codestreams_data(**kwargs):
+    """
+    Set the codestream data for the global _cs_data instance.
+
+    Args:
+        **kwargs: Key-value pairs of attributes and values to update in _cs_data.
+    """
+    global _cs_data
+    for key, value in kwargs.items():
+        assert(hasattr(_cs_data, key))
```
marcosps, line 152: I think that assert can be used without the `()`, but I'm not sure what's the more `pythonic` way to be honest...

klpbuild/klplib/config.py
```diff
@@ -3,17 +3,27 @@
...
+from pathlib import Path, PurePath
+from configparser import ConfigParser
 import logging
 import os
-from pathlib import Path, PurePath
 
+_loaded = False
+_config = ConfigParser()
+
+def check_config_is_loaded(func):
```
marcosps, line 17: Maybe you can add more information how this wrapper function stuff works. I know that we can do research about it, but it would be good to have a commit message that says how this works compared to the old method.

klpbuild/klplib/config.py
```diff
@@ -4,24 +4,9 @@
 # Author: Marcos Paulo de Souza <mpdesouza@suse.com>
 
 import configparser
-import dataclasses
```
marcosps, line 7: It would be nice if you describe in the commit message that now we call `load_codestreams` on main, and the data it loads will be used in the subsequent subcommands that use it, which is very clever. Having commits describing big changes like this is good since we may need such explanations like this in the future.

BTW, I thin that we can move this codestreams.py to klplib is a good choice here: `codestream.py` deals with the data about each codestream, while the `codestreams.py` contains the methods to handle the data.

klpbuild/klplib/config.py
```diff
@@ -24,13 +24,13 @@ def wrapper(self, *args, **kwargs):
...
-        self.user_conf_file = Path(home, ".config/klp-build/config")
-        if not self.user_conf_file.is_file():
+
+        user_conf_file = get_user_conf_file()
+        if not user_conf_file.is_file():
             logging.warning("Warning: user configuration file not found")
             # If there's no configuration file assume fresh install.
             # Prepare the system with a default environment and conf.
-            self.setup_user_env(Path(home, "klp"))
+            self.setup_user_env(Path.home()/"klp")
```
marcosps: Interesting, a new syntax for path handling. I would vote to keep the current way of path addressing, but we can discuss about it.

klpbuild/klplib/config.py
```diff
@@ -113,4 +116,6 @@ def get_tests_path(self, lp_name):
 
         raise RuntimeError(f"Couldn't find {test_sh} or {test_dir_sh}")
 
-
+def get_user_conf_file():
+    home = Path.home()
+    return home/".config/klp-build/config"
```
marcosps: I think that we can drop the home variable here.

marcosps: Also,this could be moved to utils as well.

klpbuild/klplib/utils.py
```diff
@@ -274,3 +276,15 @@ def get_kgraft_branch(cs_name):
...
+    """
+    Get the working directory for a given livepatch name.
+
+    Args:
+        lp_name (str): The name of the livepatch.
+
+    Returns:
+        Path: The full path to the livepatch file.
+    """
+    return get_user_path('work_dir')/lp_name
```
marcosps, line 292: Again, it seems to create a Path object by the syntax automatically?

klpbuild/plugins/extractor.py
```diff
@@ -482,7 +482,7 @@ def process(self, args):
             f.truncate()
 
     def run(self):
-        logging.info(f"Work directory: {self.lp_path}")
+        logging.info(f"Work directory: {utils.get_workdir(self.lp_name)}")
```
marcosps, line 484: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

I've seeing this message when using pylint, and I was about to start changing it in the future, but as you are touching some logging messages could you try to use this new pattern?

klpbuild/klplib/ibs.py
```diff
@@ -155,11 +155,17 @@ def download_cs_data(self, cs_list):
             "kernel-source": r"(kernel-(source|devel)(\-rt)?\-?[\d\.\-]+.noarch.rpm)",
         }
 
-        dest = Path(self.data, "kernel-rpms")
+        dest = Path(get_datadir(), "kernel-rpms")
```
marcosps, line 158: At this point you kept the old Path format.

klpbuild/klplib/config.py
```diff
@@ -26,8 +26,7 @@ def wrapper(*args, **kwargs):
 
 class Config:
     def __init__(self, lp_name):
-        self.data = get_user_path('data_dir')
-
+        pass
```
marcosps: Do we need a constructor if we don't have anything to set on the class at this point? Maybe you delete this in the next commits and also remove the `super` call, but this got my attention.

klpbuild/klplib/config.py
```diff
@@ -24,9 +24,6 @@ def wrapper(*args, **kwargs):
         return func(*args, **kwargs)
```
marcosps, line 34: I wasn't expecting the config class to be dropped! Cool!

marcosps:
I really liked the approach, and I only have some nitpicks about it. The code seems more organised as well. thanks a lot for doing it! I would vote to move the codestreams.py to klplib since it would be used by all plugins.

klpbuild/plugins/codestreams.py
vmezzela: Pathilib has something like:

```py
def __div__(self, other):
    return self.join(os.fspath(other))
```

I see this is [supported also in python3.5](https://docs.python.org/3.5/library/pathlib.html#operators). Do we have to support older versions as well? If so, I haven't checked yet.

klpbuild/plugins/codestreams.py
vmezzela: mmh, right! I believe the way you suggest is better:
```py
assert hasattr(_cs_data,key)
```

klpbuild/klplib/utils.py
vmezzela: Yup: `get_user_path()` return a `Path` object, and the `/` operator is overloaded to behave ad a `join()`.

klpbuild/klplib/config.py
vmezzela: mmh, in a following commit I add the docstrings for all the methods in `config` (see [here](https://github.com/SUSE/klp-build/pull/69/commits/036257f7e072da710b59634dfa055205618e3d2a#diff-18cd39e2aefd71105068b225cf4811eeca3bff3d50633ef57e1f4b2390fdc2a9R18)). I can anyway add more information in the commit body, it's still important. You're right.

I just realized that I should apply name mangling to this method as is not used outside this module and it should not be used elsewhere either.

klpbuild/klplib/config.py
vmezzela: I also thought about this solution: I was just worried that `codestream.py` and `codestramS.py` are very similar names and it can be confusing if both of them are in `klplib`.

Perhaps something like `codestream.py` and `codestreams_data.py`?

regarding the commit message: in the last iteration of this review I can amend all the commits and add more information in the changelog.

klpbuild/klplib/config.py
vmezzela: Sure, let me know if I have to revert the changes using the current way.

klpbuild/klplib/config.py
vmezzela: > I think that we can drop the home variable here.

Sure, you're right.

> Also,this could be moved to utils as well.

mmh, in a following commit this function has been renamed to `__get_user_conf_file()`, I wouldn't move it because this is an internal function of `/klplib/config.py` and it's not used elsewhere.

This module should just expose the functions `get_user_path()` and `get_user_setting()`, I would keep everything else related to how the `config` is handled hidden from outside this package.

klpbuild/plugins/extractor.py
vmezzela: Ouch, I didn't know about that. I will change it.

klpbuild/klplib/ibs.py
vmezzela: Ops, let's decide together which way to use and I will fix everything.

klpbuild/klplib/config.py
vmezzela: This was just temporary (and does not break things), in the next commit the `Config` class is totally dropped.
In this way each commit was more atomic. I found it better to proceed this way. 

Let me know if this is something I have to change.

vmezzela:
Before the next iteration, I'd like to know if you prefer:

1. to keep the old `Path().join()` or slowly introduce the new way using `/`;
2. to have `klplib/codestream.py` and `klplib/codestreamS.py` or something different;

In my opinion, regarding `1.`, maybe we can slowly introduce the new way (I like it more :satisfied:). Whereas, concerning `2.` , a solution could be having `klplib/codestream.py` and `klplib/codestreams_data.py` or `klplib/codestreams/codestream.py` and `klplib/codestreams/codestreams_data.py`. To be honest I don't know what is better.

marcosps:
> Before the next iteration, I'd like to know if you prefer:
> 
> 1. to keep the old `Path().join()` or slowly introduce the new way using `/`;

Ok, it's not something `bad` using the `/`, but it was only different, I nevere saw it before... ok, if it works, it works, and then we don't need to import pathlib for cases like this, right?

> 2. to have `klplib/codestream.py` and `klplib/codestreamS.py` or something different;
> 
> In my opinion, regarding `1.`, maybe we can slowly introduce the new way (I like it more 😆). Whereas, concerning `2.` , a solution could be having `klplib/codestream.py` and `klplib/codestreams_data.py` or `klplib/codestreams/codestream.py` and `klplib/codestreams/codestreams_data.py`. To be honest I don't know what is better.

I agree to rename codestream.py to codestream_data.py. IMHO I believe that having both codestreams_data.py and codestreams.py in klplib would be the better choise, since they are basically handling something inherent to the procedure of handling codestreams, and not some subcommand or anything else.

For the rest, I think that it's pretty good. Let's just try to be consistent regarding the paths handling (ok to use the new way whenever we change any path resolving code), and that's it!

Thanks again for working on it!

klpbuild/klplib/ibs.py
marcosps: We can use the new format is you want, np :)

klpbuild/klplib/config.py
marcosps: Ok, nothing to change, all good!

klpbuild/klplib/config.py
marcosps: np, you can use the new way.

klpbuild/klplib/config.py
marcosps: ok, amending seems nice

klpbuild/plugins/codestreams.py
marcosps: nope, we don't need to support older python versions, I just asked because I never saw this syntax before.

fgyanz (commented):
Code-wise LGTM! I'd like to test the config part since there's been some changes and we don't have automated tests yet.

klpbuild/klplib/ibs.py
```diff
@@ -23,8 +23,8 @@
 from natsort import natsorted
 from osctiny import Osc
 
-from klpbuild.klplib.config import get_user_path, get_user_settings, get_tests_path
-from klpbuild.klplib.utils import ARCH, ARCHS, get_all_symbols_from_object, get_datadir, get_elf_object, get_cs_branch, get_kgraft_branch, filter_codestreams, get_workdir
+from klpbuild.klplib.config import get_user_path
```
vmezzela, line 26: Ops, here I mistakenly dropped `get_user_settings`

vmezzela:
> Code-wise LGTM! I'd like to test the config part since there's been some changes and we don't have automated tests yet.

Sure, do you mind waiting for me to push the updated version before testing, just to avoid double work?

While updating, I spotted some (minor) mistakes.

vmezzela:
I've updated this PR with the following changes:

1. changed `Path().join()` to only use `/` in the lines touched by the changes;
2. moved `plugins/codestreams.py` to `klplib/codestreams_data.py`;
3. added commit changelog where was needed;
4. dropped `global` where it was not needed;
5. changed `hasattr()` to `hasattr`;
6. fixed minor pylint warnings;
7. applied name mangling to `check_config_is_loaded` decorator in `klplib/config.py`

fgyanz (approved):
LGTM!

PR #70, opened 2025-01-24
tox: Use the ssh-agent from host

fgyanz:
Fetch the ssh-keys from the ssh-agent running in the host. Needed when klp-build needs to download new codestreams from IBS during the tests.

Otherwise we get:
```
tests/test_config.py::test_filter PASSED                                                                                                                                              [  5%]
tests/test_extract.py::test_detect_file_without_ftrace_support Enter passphrase for "/home/fgonzalez/.ssh/suse-fgonzalez":
```

A better solution would be to disable downloading codestreams during the testing, but that'd require quite some changes in setup. As a temporal workaround this does the job.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@ skip_missing_interpreters = True
 minversion = 3.28.0
 
 [testenv:{py36,py38,py310,py311,py312}-tests,tests]
+passenv = SSH_AUTH_SOCK
 description = run tests
 deps =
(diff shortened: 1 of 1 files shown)


marcosps (approved):
I think that we should remove the Makefile that can also start pytest, but tox should be the tool to do that. If this fixes your issue, than we are good :)

vmezzela (approved):

PR #75, opened 2025-01-30
Drop bsc_check from scan

vmezzela:
Before this patch, executing the `scan` command would create a livepatch working directory, even though it should not setup any workspace (which is instead done by the `setup` command).

There are two reasons behind this misbheavior:

1.  `scan` is a method of the `GitHelper` class, which includes `lp_name` as one of its fields. While this fiels is irrelevant for scan, it was still required, forcing us to provide a dummy value just to instantiate the GitHelper object. See `main.py`:
    ```py
    elif args.cmd == "scan":
            GitHelper("bsc_check", "", "").scan(args.cve, args.conf, False)
    ```
    Here, `bsc_check` was used as a placeholder `lp_name` just to allow instantiation.

2. `scan` relied on `get_commits` (the function behind the `get-patches` command), which did more than just retrieving fixes and upstream commits. It also stored them in the livepatch workdir, effectively creating an implicit `bsc_check` directory.

Furthermore, `scan` uses the function `download_supported_file`, which requires the `lp_name` (for the `lp_path`) even though its job is just to retrieve the supported codestreams, a thing that doesn't require and it's not related at all to the `lp_name` and the `lp_path` of a specific livepatch.

The reason why this is required is that the function `download_supported_file`, beside downloading the supported file, also parses it and return the supported codestreams in the form of a list of `Codestream` objects, which among their fields also have the `lp_path`, `lp_name` and `data_path`.

These three fields are not strictly required in the `Codestream` class, because they can be retrieved at runtime from the config file and from the current livepatch name.

So to ensure the `scan` command does not create any working directory, I've first dropped `data_path`, `lp_path` and `lp_name` fields from the `Codestream` class and from the `download_supported_file` function (which has also been moved in a dedicated module in klplib since this is not strictly related to `ksrc` which should only concern kernel sources).

Coming back to the issue `1`, I've dropped `lp_name` field from the `GitHelper` so it is no longer required when instantiating the class. To solve issue `2` I made `get_commits` capable of storing patches only when explicitly requested.

Now, scan can do its job without requiring a working directory.

While working on it, I renamed some variables and functions here and there to help me reading the code.

I believe there are other places where we should drop unnecessary fields: for instance, when dropping "bsc_check", I had to write something like:
```py
IBS("", self.lp_filter).download_cs_data(data_missing)
```
Here with `IBS` it's happening the same problem as with `GitHelper`, but that's surely an issue for another Pull Request.

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -140,17 +140,23 @@ class Codestream:
         return f"{self.sle}.{self.sp}u{self.update}"
 
 
-    def dir(self):
+    def get_ccp_dir(self):
+        """
+        Get the path to the ccp directory of the current codestream.
+
+        Returns:
+            Path: The path to the ccp directory of the current codestream.
+        """
         return Path(self.lp_path, "ccp", self.name())
 
 
     def lpdir(self):
-        return self.dir()/"lp"
+        return self.get_ccp_dir()/"lp"
 
 
     def work_dir(self, fname):
         fpath = f'work_{str(fname).replace("/", "_")}'
-        return Path(self.dir(), fpath)
+        return self.get_ccp_dir()/fpath
 
 
     def name_cs(self):
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -150,7 +150,13 @@ class Codestream:
         return Path(self.lp_path, "ccp", self.name())
 
 
-    def lpdir(self):
+    def get_lp_dir(self):
+        """
+        Get the path to the extracted livepatches directory of the current codestream.
+
+        returns:
+            path: The path to the extracted livepatches directory of the current codestream.
+        """
         return self.get_ccp_dir()/"lp"
 
 
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -57,6 +57,8 @@ class Codestream:
     @classmethod
     def from_cs(cls, cs):
         match = re.search(r"(\d+)\.(\d+)(rt)?u(\d+)", cs)
+        if not match:
+            raise ValueError("Filter regexp error!")
         return cls("", "", int(match.group(1)), int(match.group(2)),
                    int(match.group(4)), match.group(3), "", "", "", [], {}, {})
 
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -45,10 +45,11 @@ class Codestream:
             sle, _, u = cs.replace("SLE", "").replace("-RT", "").split("_")
             if "-SP" in sle:
                 sle, sp = sle.split("-SP")
-
         # MICRO-6-0_Update_2
         elif "MICRO" in cs:
             sle, sp, u = cs.replace("MICRO-", "").replace("-RT", "").replace("_Update_", "-").split("-")
+        else:
+            assert False, "codestream name should contain either SLE or MICRO!"
 
         return cls(data_path, lp_path, int(sle), int(sp), int(u), rt, proj, patchid, kernel, [], {}, {})
 
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -6,18 +6,15 @@
 from pathlib import Path, PurePath
 import re
 
-from klpbuild.klplib.utils import ARCH, is_mod, get_all_symbols_from_object, get_datadir
+from klpbuild.klplib.utils import ARCH, get_workdir, is_mod, get_all_symbols_from_object, get_datadir
 
 class Codestream:
-    __slots__ = ("data_path", "lp_path", "lp_name", "sle", "sp", "update", "rt",
-                 "ktype", "needs_ibt", "is_micro", "project", "patchid", "kernel",
-                 "archs", "files", "modules", "repo")
+    __slots__ = ("sle", "sp", "update", "rt", "ktype", "needs_ibt", "is_micro",
+                 "project", "patchid", "kernel", "archs", "files", "modules",
+                 "repo")
 
-    def __init__(self, data_path, lp_path, sle, sp, update, rt, project,
-                 patchid, kernel, archs, files, modules):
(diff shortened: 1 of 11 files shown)


klpbuild/klplib/utils.py
```diff
@@ -292,14 +292,16 @@ def get_workdir(lp_name):
...
-def get_datadir():
+def get_datadir(arch=""):
     """
     Get the data directory.
 
     Returns:
         Path: The full path to the livepatch file.
     """
-    return get_user_path('data_dir')
+    if arch: assert arch in ARCHS
```
marcosps, line 302: typo? the indentation seems weird :)

klpbuild/klplib/codestream.py
```diff
@@ -45,10 +45,11 @@ def from_codestream(cls, data_path, lp_path, cs, proj, patchid, kernel):
             sle, _, u = cs.replace("SLE", "").replace("-RT", "").split("_")
             if "-SP" in sle:
                 sle, sp = sle.split("-SP")
-
         # MICRO-6-0_Update_2
         elif "MICRO" in cs:
             sle, sp, u = cs.replace("MICRO-", "").replace("-RT", "").replace("_Update_", "-").split("-")
+        else:
+            raise ValueError("codestream name should contain either SLE or MICRO!")
```
marcosps, line 49: maybe an assert would do it?

klpbuild/klplib/supported.py
```diff
@@ -0,0 +1,55 @@
...
+#
+# Copyright (C) 2021-2025 SUSE
+# Author: Marcos Paulo de Souza <mpdesouza@suse.com
+
+import logging
+import re
+import requests
+from pathlib import Path
+
+from klpbuild.klplib.codestream import Codestream
```
marcosps, line 11: Good idea! Makes much more sense to create a new new module!

marcosps:
I agree with the changes, but I haven't checked if the tests pass on my local machine. Btw, I believe that you could also rebase on top of my PR that removed the `skip` argument, since it's already removed there. besides, good work! Things got much simpler now!

klpbuild/klplib/utils.py
vmezzela: It's rarely used but you can write `if` statements in one line.
I only use it when the statement is very short.

Otherwise I can also use:
```suggestion
    assert arch in ARCHS if arch else True
```

or the normal way
```suggestion
    if arch: 
        assert arch in ARCHS
```

Let me know which one you prefer :smiley:

klpbuild/klplib/codestream.py
vmezzela: That's true, in this case an assert is better. I'll change it!

vmezzela:
All the tests are passing on my machine, besides that I'm currently using this branch on `klp-build` for my next livepatch so I do further testing. After that I'll mark this PR ready to be merged.

I will rebase it on top #73 (and/or #74) when it gets merged :rocket:

klpbuild/klplib/utils.py
marcosps: Personally I would use the second one :)

vmezzela: Sure, will change it :smiley:

vmezzela:
Rebased on current `devel`.

klpbuild/klplib/codestream.py
```diff
@@ -140,17 +140,23 @@ def name(self):
         return f"{self.sle}.{self.sp}u{self.update}"
 
 
-    def dir(self):
+    def get_cpp_dir(self):
```
marcosps: I think that it should be named `ccp`, right? :)

klpbuild/klplib/codestream.py
```diff
@@ -40,8 +40,6 @@ def from_codestream(cls, data_path, lp_path, cs, proj, patchid, kernel):
         sp = "0"
         u = "0"
 
-        assert "SLE" in cs or "MICRO" in cs
```
marcosps: Merge this commit with the previous one?

klpbuild/klplib/ksrc.py
```diff
@@ -475,7 +475,7 @@ def scan(self, lp_name, cve, conf, no_check):
         if data_missing:
             logging.info("Download the necessary data from the following codestreams:")
             logging.info("\t%s\n", " ".join(cs_missing))
-            IBS(lp_name, self.lp_filter).download_cs_data(data_missing)
+            IBS("", self.lp_filter).download_cs_data(data_missing)
```
marcosps, line 478: Not something to work on now, but I'm seeing that maybe we could avoid having lp_name on IBS object as well, at least for download_cs_data. Maybe this could be moved into utils as well, but this function call `get_cs_packages`, which then uses the self.osc.

Well, maybe one day we can decouple things like this :)

marcosps:
Just small nits, awesome work. Thanks a lot! With those fixed (mostly the cpp vs ccp one), you are free to merge it!

klpbuild/klplib/codestream.py
vmezzela: oh right! :laughing:

klpbuild/klplib/codestream.py
vmezzela: I must have done something wrong while rebasing, I'll fix it

klpbuild/klplib/ksrc.py
vmezzela: I think we can keep `get_cs_package` in `klplib/ibs.py`, but definitely we can refactor the IBS class (does it need to be a class?) and remove unused attributes, like `lp_name`.

We  can do that while converting this module into a plugin.

marcosps: Agreed. We can start move some code out of the IBS class, since it doesn't need to be in a class at all. But as I said, this is for the future us :)

marcosps (approved):
LGTM!

fgyanz:
LGTM too! Code reviewed, all the changes make sense. Also tested and no problems were found.

fgyanz (approved):

PR #78, opened 2025-02-05
Introduce plugin system

vmezzela:
This Pull Request introduces the first version of a plugin system, aiming to modularize all commands into standalone plugins.

## Plugin Development Overview

A plugin is nothing else than just a module in the package `klpbuild.plugins` that exposes (at least) two functions:

1. `register_argparser()` : 
    - This function registers the plugin’s command-line arguments.
    - Get the parent parser in the arguments and should add a subparser to it.

3. `run()` : 
    - This function executes the plugin's logic.
    -  It should accept arguments that were registered using the above function `register_argparser()`

    e.g.:
    ```python3
    def run(lp_name, lp_filter):
        pass
    ```

When introducing a plugin, there's no need to modify `klpbuild/main.py` as the plugins are detected at runtime if they have the above mentioned `run` function.

## Internal Workflow
Here’s how the new system processes commands in chronological order:

1. Building the Argument Parser:
    - `create_parser` is called from the main to build the argument parser.
    - `register_plugins_argparser` is called from `create_parser`: this function iterate over all the modules in `klpbuild.plugins` and calls all the `register_argparser()` of each plugin, registering all the subparser in the main parent parser.

2. Executing a plugin:
    - `try_run_plugin` is called from main, it fetches the plugin to run from `klpbuild.plugins` and executes its `run()` function.
    - Before running the plugin, `try_run_plugin` extracts from the parsed arguments only those that concerns this plugin (i.e. those registered via `register_argparser()` and passes them to `run()`.

--- 

With this Pull Request, I also convert the command `scan` to a plugin to give an idea of how the system works. 

Currently, the codebase is hybrid as all the other commands still work in the old way (they are just functions somewhere). As a consequence, in this current state the codebase is messy: the changes required in `klplib/cmd.py` to introduce the per-plugin subparser introduced repetitive code in the function `create_parser`, but it'll be gone when all the command will be converted into plugins. Same for all the if-branches in the main function.

I'm not hundred percent sure this is a good way to go about this plugin subsystem and I'm open to any feedback, please let me know if you would change anything or totally use another solution.

I'm marking this PR as draft: before merging I want to try to convert other commands into plugins to make sure this system works smoothly and without issues, but meanwhile I still appreciate you feedback on this.

### DIFF
--- a/klpbuild/main.py
+++ b/klpbuild/main.py
@@ -38,42 +38,42 @@ def main():
         ffuncs = Setup.setup_file_funcs(args.conf, args.module, args.file_funcs,
                                         args.mod_file_funcs, args.conf_mod_file_funcs)
         codestreams = setup.setup_codestreams(
-            {"cve": args.cve, "conf": args.conf, "lp_filter": args.filter,
+            {"cve": args.cve, "conf": args.conf, "lp_filter": args.lp_filter,
                 "no_check": args.no_check})
         setup.setup_project_files(codestreams, ffuncs, args.archs)
 
     elif args.cmd == "extract":
-        Extractor(args.name, args.filter, args.apply_patches, args.avoid_ext).run()
+        Extractor(args.name, args.lp_filter, args.apply_patches, args.avoid_ext).run()
 
     elif args.cmd == "cs-diff":
-        Extractor(args.name, args.filter, False, []).cs_diff()
+        Extractor(args.name, args.lp_filter, False, []).cs_diff()
 
     elif args.cmd == "check-inline":
         Inliner(args.name, args.codestream).check_inline(args.file, args.symbol)
 
     elif args.cmd == "get-patches":
-        GitHelper(args.filter).get_commits(args.cve, get_workdir(args.name))
+        GitHelper(args.lp_filter).get_commits(args.cve, get_workdir(args.name))
 
     elif args.cmd == "scan":
         GitHelper("").scan(args.cve, args.conf, False)
 
     elif args.cmd == "format-patches":
-        GitHelper(args.filter).format_patches(args.name, args.version)
+        GitHelper(args.lp_filter).format_patches(args.name, args.version)
 
     elif args.cmd == "status":
-        IBS(args.name, args.filter).status(args.wait)
+        IBS(args.name, args.lp_filter).status(args.wait)
 
     elif args.cmd == "push":
-        IBS(args.name, args.filter).push(args.wait)
+        IBS(args.name, args.lp_filter).push(args.wait)
 
     elif args.cmd == "log":
-        IBS(args.name, args.filter).log(args.arch)
+        IBS(args.name, args.lp_filter).log(args.arch)
 
     elif args.cmd == "cleanup":
-        IBS(args.name, args.filter).cleanup()
+        IBS(args.name, args.lp_filter).cleanup()
 
     elif args.cmd == "prepare-tests":
-        IBS(args.name, args.filter).prepare_tests()
+        IBS(args.name, args.lp_filter).prepare_tests()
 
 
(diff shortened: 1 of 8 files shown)


klpbuild/main.py
```diff
@@ -24,6 +25,14 @@ def main():
     if hasattr(args, 'name'):
         load_codestreams(args.name)
 
+    try:
+        try_run_plugin(args.cmd, args)
```
marcosps, line 29: We don't check if the args.cmd is a plugin in the first place, so it fails when trying to run commands other than `scan`:

```
$ klp-build cs-diff --name bsc1231993 --filter "15.5u9|15.6u[0]"
Loading user configuration from '/home/mpdesouza/.config/klp-build/config'
  File "/home/mpdesouza/.local/bin/klp-build", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/mpdesouza/.local/pipx/venvs/klp-build/lib64/python3.11/site-packages/klpbuild/main.py", line 29, in main
    try_run_plugin(args.cmd, args)
  File "/home/mpdesouza/.local/pipx/venvs/klp-build/lib64/python3.11/site-packages/klpbuild/klplib/plugins.py", line 32, in try_run_plugin
    module = __get_plugin(name)
             ^^^^^^^^^^^^^^^^^^
  File "/home/mpdesouza/.local/pipx/venvs/klp-build/lib64/python3.11/site-packages/klpbuild/klplib/plugins.py", line 90, in __get_plugin
    return importlib.import_module(PLUGINS_PATH + name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'klpbuild.plugins.cs-diff'

```

vmezzela: right, I messed this up this morning on a last minute change.. :laughing:

marcosps:
@vmezzela thanks for fixing the problem that I found! OTHO, would you please try to address the pylint complaints for newly created files? Some of the reported issues already existed on the old code as well, but since we are moving code around it's easier to address them.

marcosps (approved):
LGTM!

vmezzela:
Hi @marcosps , I removed the pylint warning related to my PR. Some of them come from the `scan` function, it might need to be reworked to get rid of those. Maybe in another PR.

While there's still one:
```
Redefining name 'scan' from outer scope (line 36) (redefined-outer-name)
```
I guess we have it because there's the `scan` function that has the same name as its own module, but I'd keep it rather than renaming because of this warning. 

What do you think?

marcosps:
> Hi @marcosps , I removed the pylint warning related to my PR. Some of them come from the `scan` function, it might need to be reworked to get rid of those. Maybe in another PR.
> 
> While there's still one:
> 
> ```
> Redefining name 'scan' from outer scope (line 36) (redefined-outer-name)
> ```
> 
> I guess we have it because there's the `scan` function that has the same name as its own module, but I'd keep it rather than renaming because of this warning.
> 
> What do you think?

np, let these be there until we finish the plugins conversion. I believe that we'll have a slightly different code by then, and many of these warnings would be fixed anyway.

fgyanz (approved):
LGTM!

PR #80, opened 2025-02-10
Improve ksrc heuristic yet again

### LINKED ISSUE #68: Restrict kernel patches search space
There's a corner case where `is_kernel_patched()` in `GitHelper` can mistakenly grab a commit that does not concern any patch. 
The false-positive commit will result in a codestream being reported as not affected, while it actually is because the actual patch is not backported for the codestream in question.

Using `git log --grep=CVE-{cve} -- patches.suse` instead of `git log --grep=CVE-{cve}` would filter out those commits unrelated to patches.

The following code would solve the problem:
```diff
@@ -343,7 +344,10 @@ class GitHelper(Config):
         ret = subprocess.check_output(["/usr/bin/git", "-C", self.kern_src, "log",
                                        f"--grep=CVE-{cve}",
                                        f"--tags=*rpm-{kernel}",
-                                       "--pretty=oneline"])
+                                       "--pretty=oneline",
+                                       "--",
+                                       "patches.suse"])
```
But it slows down a lot the execution of `git log` because Git doesn't store which files are changed in each commit.

What we could do instead, is to keep `git log` as it is, and filter out commits unrelated to `patches.suse` only aftewards.

A cve that tigger this issue is `2024-36979`.

fgyanz:
Hi team,

This PR improves some parts of the kernel-sources heuristic:

- (`get_commits`) Skip commits that modify the patch file but without affecting the diff part. Most times such commits only change the patch's metadata and surrounding text.

- (`is_kernel_patched`) Only analyze commits that are actually related to the CVE. Skip commits that do not match 1:1 with the ones found in the SLE branch. Fixes #68 and some other issues.

Tested with:

- CVE `2021-47511`. klp-build was incorrectly reporting `15.3u40-44` as vulnerable.
- CVE `2024-36979`. klp-build was incorrectly reporting `15.6u3 and 15.6rtu1` as patched.
- CVE `2024-45016`. A rather big and complex CVE fix. Still works fine.

No noticeable slowdown was appreciated during testing.

### DIFF
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -133,6 +133,34 @@ class GitHelper():
                                 list(self.kernel_branches.values()))
 
 
+    def diff_commits(self, base, new, patch):
+        """
+        Check if there's any difference between the two given commits.
+
+        Args:
+            base (str): Commit to use as base in the diff.
+            new (str): Commit to compare with the base.
+            patch (list): Target files in the diff.
+
+        returns:
+            Boolean: True if both commits differ. False otherwise.
+        """
+
+        if base == "":
+            return True
+
+        # Compare lines starting with '+' or '-'.
+        # This should be enough to ignore sneaky metadata updates on
+        # the file and duplicated commits.
+        diff = subprocess.run(["/usr/bin/git", "-C", str(self.kern_src), "diff",
+                               "--numstat", r"-G'^\+|^-'", base, new,
+                               "--", str(patch)],
+                              stdout = subprocess.DEVNULL,
+                              stderr = subprocess.DEVNULL)
+
+        return diff.returncode
+
+
     def get_commits(self, cve, savedir=None):
         if not self.kern_src:
             logging.info("kernel_src_dir not found, skip getting SUSE commits")
@@ -243,6 +271,7 @@ class GitHelper():
                             self.kern_src,
                             "log",
                             "--numstat",
+                            "--reverse",
                             "--no-merges",
                             "--pretty=oneline",
                             f"remotes/origin/{mbranch}",
@@ -260,6 +289,7 @@ class GitHelper():
                     continue
 
                 iphashes = iter(phashes.splitlines())
+                base = ""
                 for hash_entry in iphashes:
                     stats = next(iphashes)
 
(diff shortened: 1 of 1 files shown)


klpbuild/klplib/ksrc.py
```diff
@@ -384,15 +403,29 @@ def is_kernel_patched(self, kernel, suse_commits, cve):
         ret = subprocess.check_output(["/usr/bin/git", "-C", self.kern_src, "log",
                                        f"--grep=CVE-{cve}",
                                        f"--tags=*rpm-{kernel}",
-                                       "--pretty=oneline"])
+                                       "--format='%at-%H-%f'"]).decode().splitlines()
+        # Sort by date
+        ret.sort(reverse=True)
```
marcosps, line 408: In this case can't we use `--reverse` as we are using `git log` here?

fgyanz: Unfortunately `git-log` doesn't promise that it will order the commits by date. It will try, but you can't take it for granted (e.g. CVE  `2024-45016`). So what I do is:
-  Print the commits in the format: `{date-unix-format}-{commit}-{commit-title}`
- And then sort the list by date.

klpbuild/klplib/ksrc.py
```diff
@@ -384,15 +403,29 @@ def is_kernel_patched(self, kernel, suse_commits, cve):
...
+            c = line.split("-")[1]
+
+            files = self.get_commit_files(c, r"patches\.suse\/.+\.patch")
+            if len(files) == 0:
+                continue
+
+            # Match 1:1 with the commits found in SLE branch
+            for s in suse_commits:
+                diff = self.diff_commits(s, c, files)
+                if diff == False:
```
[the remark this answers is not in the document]
fgyanz: done

marcosps:
Ok, I do not completely understand how things really work here anymore (it has been some much since I started this code), but it somehow makes sense when reading the code. Thanks a lot for working on it!

marcosps (approved):
LGTM!

vmezzela (approved):
I've tested it on another livepatch and it worked fine. I also like the changes very much! Thanks! :rocket:

PR #83, opened 2025-02-12
codestream: Test if cs.find_obj_path respects the arch passed as argu…

marcosps:
…ment

The code currently handles a cache to not skip searching the same object path when searching for multiple symbols on different files on the same kernel version.

In this case we can skip the cache if the arch is different from ARCH, which is the one used when creating the cache in the first place.

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -271,7 +271,8 @@ class Codestream:
 
     # Return only the name of the module to be livepatched
     def find_obj_path(self, arch, mod):
-        # Return the path is the modules was previously found
+        # Return the path if the modules was previously found for ARCH, or refetch if
+        # the obejct is for a different architecture
         obj = self.modules.get(mod, "")
         if obj:
             assert self.kernel in str(obj)
@@ -279,7 +280,7 @@ class Codestream:
 
         # We already know the path to vmlinux, so return it
         if not is_mod(mod):
-            return self.get_boot_file("vmlinux", arch)
+            return self.get_boot_file("vmlinux", arch).relative_to(get_datadir(arch))
 
         # Module name use underscores, but the final module object uses hyphens.
         mod = mod.replace("_", "[-_]")
@@ -298,7 +299,7 @@ class Codestream:
             if Path(mod_path, obj).exists():
                 break
 
-        return Path(mod_path, obj)
+        return Path(mod_path, obj).relative_to(get_datadir(arch))
 
 
     def lp_out_file(self, lp_name, fname):
@@ -316,7 +317,7 @@ class Codestream:
         cache[arch].setdefault(name, {})
 
         if not cache[arch][name].get(mod, ""):
-            obj = self.find_obj_path(arch, mod)
+            obj = get_datadir(arch)/self.find_obj_path(arch, mod)
             cache[arch][name][mod] = get_all_symbols_from_object(obj, True)
 
         ret = []
--- a/klpbuild/klplib/utils.py
+++ b/klpbuild/klplib/utils.py
@@ -16,7 +16,7 @@ import zstandard
 from elftools.common.utils import bytes2str
 from elftools.elf.elffile import ELFFile
 from elftools.elf.sections import SymbolTableSection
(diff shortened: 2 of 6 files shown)


marcosps:
@fgyanz @vmezzela this is the bug that I told you today :)

vmezzela (commented):
If I understand correctly, what happens is that:

1. during `setup_project_files`, we save in `Codestream.modules` the path of the module only for the current ARCH:
    ```py
                mod_path = cs.find_obj_path(utils.ARCH, mod)

                # Validate if the module being livepatched is supported or not
                if utils.check_module_unsupported(mod_path):
                    logging.warning("%s (%s}): Module %s is not supported by SLE", cs.name(), cs.kernel, mod)

                cs.modules[mod] = str(mod_path)
    ```
2. This patch make sure that when call `find_obj_path`, it uses the value saved beforehand only if the arch match the current arch (i.e. `utils.ARCH`), and this thankfully solves the bug. :rocket: 

My question is, could we save the path for the other archs as well? or even, if `find_obj_path` is not called often we could also think about dropping the `modules` field in the `Codestream` class if it's used only for this use case. Not sure though this suggestions make sense, I haven't looked into the details of it.

fgyanz (commented):
Hi @marcosps . Do you mind sharing the CVE that triggered this bug? I imagine the testcase is based on it, but I also want  test it manually. Thanks!
Also, would it be possible to change the target branch to master? :sweat_smile:

marcosps:
> If I understand correctly, what happens is that:
> 
> 1. during `setup_project_files`, we save in `Codestream.modules` the path of the module only for the current ARCH:
>    ```python
>                mod_path = cs.find_obj_path(utils.ARCH, mod)
>    
>                # Validate if the module being livepatched is supported or not
>                if utils.check_module_unsupported(mod_path):
>                    logging.warning("%s (%s}): Module %s is not supported by SLE", cs.name(), cs.kernel, mod)
>    
>                cs.modules[mod] = str(mod_path)
>    ```
> 2. This patch make sure that when call `find_obj_path`, it uses the value saved beforehand only if the arch match the current arch (i.e. `utils.ARCH`), and this thankfully solves the bug. 🚀
> 
> My question is, could we save the path for the other archs as well? or even, if `find_obj_path` is not called often we could also think about dropping the `modules` field in the `Codestream` class if it's used only for this use case. Not sure though this suggestions make sense, I haven't looked into the details of it.

Ok, think about about a livepatch that has a couple files targgeting the same module. For for each file that we are going to extract symbol we would need to parse the modules.order again and again to search for the module path.

What usually happen is that we only create a livepatch targgeting one single module, so in this case maybe we could optmize things further: grab all symbols to validate for one object and do it only once. That would make it more clean I think.

About dropping the module path, what we could do would be to just save the path for the module since usr/lib/modules, and leave the `datadir` part to be filled by the find_obj_path per arch. That would be cleaner.

marcosps:
> Hi @marcosps . Do you mind sharing the CVE that triggered this bug? I imagine the testcase is based on it, but I also want test it manually. Thanks! Also, would it be possible to change the target branch to master? 😅

Sure, bsc1227371, CVE-2024-36974. The problem lies in symbol `taprio_parse_clockid`, which exists on x86_64 but doens't on s390x. So the LP is built, the checks against the sch_taprio are done only on x86_64, but once the LP is loaded on s390x it fails to find the symbol, and a WARN is triggered.

marcosps:
@vmezzela @fgyanz I reorganized the code to only cache the path for the module without the `datadir`, and I adjusted all the places that require the absolute path to append `datadir`. Can you please take a look? It passed the tests in my machine.

marcosps:
@fgyanz I thought about basing it on main, but TBH devel changed a lot since our last sync right? Maybe we can leave this on devel and later (maybe soon), we can merge it on main again :)

vmezzela:
Right, we didn't need to save the absolute path. And also, if the path of the given object relative to `data_dir` is the same for all the archs this would solve the issue and make the things more clean. I like it!

fgyanz:
LGTM!

PR #85, opened 2025-02-14
Add verbose global option

vmezzela:
I believe it's useful to have a global option to enable/disable debugging logs.

I open this a Draft PR so that if you have any other logging call site that would like to convert from `logging.info` to `logging.debug` we can do it in this PR and then merge it.

For instance, I propose to change it when loading the `config` file.

### DIFF
--- a/klp-build.1
+++ b/klp-build.1
@@ -24,6 +24,9 @@ Generic options available for all commands:
 .B "-h, --help"
 Show command's help message and exit.
 .TP
+.B "-v, --verbose"
+Produce more verbose output.
+.TP
 .BI "-n, --name" " NAME"
 The livepatch name. This will be the directory name of the resulting
 livepatches.
--- a/klpbuild/klplib/cmd.py
+++ b/klpbuild/klplib/cmd.py
@@ -33,6 +33,13 @@ def create_parser() -> argparse.ArgumentParser:
     parentparser = argparse.ArgumentParser(add_help=True)
     sub = parentparser.add_subparsers(dest="cmd")
 
+    parentparser.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="Produce more verbose output"
+    )
+
     register_plugins_argparser(sub)
 
     # NOTE: all the code below should be gone when all the module will be
--- a/klpbuild/klplib/cmd.py
+++ b/klpbuild/klplib/cmd.py
@@ -30,7 +30,7 @@ def add_arg_lp_filter(parentparser, mandatory=False):
 
 
 def create_parser() -> argparse.ArgumentParser:
-    parentparser = argparse.ArgumentParser(add_help=False)
+    parentparser = argparse.ArgumentParser(add_help=True)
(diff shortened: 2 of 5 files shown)


marcosps:
Interesting! Maybe we could also add more messages on scan command, since it currently shows all commits for every supported codestream? I remember that @fgyanz likes the messages, but maybe he changed his idea :P

I tried locally and it shows even messages about requests lib getting the list of supported codestreams. I would say that this is good enough for now, and then we can start moving some messages to debug as we find suitable.

marcosps:
Also, we still have some spread `print` statements that we can maybe check wether they belong to INFO or debug.

fgyanz:
> Interesting! Maybe we could also add more messages on scan command, since it currently shows all commits for every supported codestream? I remember that @fgyanz likes the messages, but maybe he changed his idea :P
> 

Agreed! I do personally find useful them useful catching bugs in the commits heuristic. But understandably it's not needed every time :)

marcosps:
> > Interesting! Maybe we could also add more messages on scan command, since it currently shows all commits for every supported codestream? I remember that @fgyanz likes the messages, but maybe he changed his idea :P
> 
> Agreed! I do personally find them useful for catching bugs in the commits heuristic. But understandably it's not needed every time :)

That's true. But given your last commits around the commits heuristic I would say that we are in a much better place now :)

fgyanz:
BTW @vmezzela, don't forget to update the man page with the new verbose command :P

vmezzela:
> BTW @vmezzela, don't forget to update the man page with the new verbose command :P

Oh right, I'll do that for sure before merging!

I'll keep this PR open a little longer in case any other changes come to mind.

marcosps (approved):
LGTM!

fgyanz:
FWIW the `--verbose` option can only be used before the klp-build's commands. For instance:
```
$ klp-build scan --cve 2022-48911 -v
usage: klp-build [-h] [-v]
                 {scan,setup,check-inline,extract,cs-diff,format-patches,get-patches,cleanup,prepare-tests,push,status,log}
                 ...
```
`--name` and `--filter` don't have such restriction. Perhaps we could have the same behavior with `verbose`?

fgyanz:
Note that `help` also behaves like `name` and `filter`, so `verbose` would actually be the sole exception that has to go before the command. I personally prefer having some consistency between all the options and commands, so that the user doesn't have to think twice before writing the command line. 

Then again, if it requires too much additional work maybe it's not worth it. But this topic will come back the next time we add a new option.

vmezzela:
Mhh okay, then I was wrong regarding the global option being commonly used before the command. :smile: 
If this is not the case, I agree with you that we should be consistent with the other options. I'll work on this as soon as I can.

fgyanz (approved):
Apparently it is not possible to have `verbose` before and after the commands with the current solution and it would require a mayor re-work. So, disregard my previous comment. The PR does the job as it is. Thanks a lot! LGTM!

PR #87, opened 2025-02-17
Bz integration

marcosps:
@vmezzela @fgyanz this is the first part of the bugzilla integration that I made to work based on @fgyanz 's https://github.com/fgyanz/klp-bugzilla.

The idea is to grab the CVE info from the bsc itself in the future, and also pave the way to integrate the automatic scan later on.

### DIFF
--- a/klpbuild/klplib/bugzilla.py
+++ b/klpbuild/klplib/bugzilla.py
@@ -1,0 +1,30 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2021-2025 SUSE
+# Authors: Fernando Gonzalez
+#          Marcos Paulo de Souza <mpdesouza@suse.com>
+
+import bugzilla
+from klpbuild.klplib.config import get_user_settings
+
+
+def get_bug_data(bsc):
+    bz_key = get_user_settings("bugzilla_api_key", True)
+    if 'changeit' in bz_key:
+        raise RuntimeError("Change the bugzilla_api_key to a valid key.")
+
+    bzapi = bugzilla.Bugzilla("https://bugzilla.suse.com", api_key=bz_key)
+
+    return bzapi.getbug(bsc)
+
+
+def get_bug_title(bsc):
+    """
+    The bug description usually is after the "kernel live patch" message. Return a fixes message
+    if the livepatch name is not an ID from a bug on bugzilla.
+    """
+    if not bsc.isnumeric():
+        return "Change me!"
+
+    data = str(get_bug_data(int(bsc)).summary)
+    return data.split("kernel live patch: ")[1].strip()
--- a/klpbuild/klplib/config.py
+++ b/klpbuild/klplib/config.py
@@ -124,7 +124,8 @@ def __setup_user_env(basedir):
                        '#kgr_patches_tests_dir': 'kgraft-patches_testscripts/',
                        '#kernel_src_dir': 'kernel-src/'}
 
-    config['Settings'] = {'workers': 4}
+    config['Settings'] = {'workers': 4,
+                          'bugzilla_api_key': 'changeit'}
 
     logging.info("Creating default user configuration: '%s'", user_conf_file)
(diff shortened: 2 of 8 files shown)


klpbuild/klplib/templ.py
```diff
@@ -565,21 +566,23 @@ def __create_kbuild(self, cs):
...
-            "lp_name": self.lp_name.replace("bsc", ""),
-            "user": user,
-            "email": email,
-            "cve": cve,
-            "commits": cmts,
-            "msg": "Upstream commits" if len(cmts) > 1 else "Upstream commit",
-        }
-        with open(get_workdir(self.lp_name)/"commit.msg", "w") as f:
-            f.write(Template(TEMPL_COMMIT).render(**render_vars))
+
```
vmezzela, line 569: Oh thanks for moving this out of the class! I think it's better!

vmezzela (changes requested):
Hi, I would approve the changes, I tested it manually and it works. But unfortunately it breaks some tests in `get_lp_name` because the livepatch name is not an int in this case.

```
    def get_lp_number(lp_name):
>       return int(lp_name.replace("bsc", ""))
E       ValueError: invalid literal for int() with base 10: '_test_templ_kbuild_has_contents'
```

marcosps:
> Hi, I would approve the changes, I tested it manually and it works. But unfortunately it breaks some tests in `get_lp_name` because the livepatch name is not an int in this case.
> 
> ```
>     def get_lp_number(lp_name):
> >       return int(lp_name.replace("bsc", ""))
> E       ValueError: invalid literal for int() with base 10: '_test_templ_kbuild_has_contents'
> ```

Ouch... I forgot to rerun the tests at this point. Thanks a lot for catching this mistake!

marcosps:
@vmezzela fixed in the last push!

klpbuild/klplib/bugzilla.py
```diff
@@ -0,0 +1,30 @@
...
+
+def get_bug_title(bsc):
+    """
+    The bug description usually is after the "kernel live patch" message. Return a fixes message
+    if the livepatch name is not an ID from a bug on bugzilla.
+    """
+    if not bsc.isnumeric():
+        return "Change me!"
+
+    data = str(get_bug_data(int(bsc)))
```
fgyanz, line 29: small nitpick, but can't we limit the scope to just `summary` field? Such as:  
`data = get_bug_data(int(bsc)).summary`
We would still need to parse it though.

klpbuild/klplib/config.py
```diff
@@ -124,7 +124,8 @@ def __setup_user_env(basedir):
                        '#kgr_patches_tests_dir': 'kgraft-patches_testscripts/',
                        '#kernel_src_dir': 'kernel-src/'}
 
-    config['Settings'] = {'workers': 4}
+    config['Settings'] = {'workers': 4,
+                            'bugzilla_api_key': 'changeit'}
```
fgyanz, line 128: can you align both strings? :)

fgyanz (changes requested):

klpbuild/klplib/config.py
marcosps: Makes sense. I didn't checked pylint in this case, sorry about it =/

klpbuild/klplib/bugzilla.py
```diff
@@ -0,0 +1,30 @@
...
+
+    return bzapi.getbug(bsc)
+
+
+def get_bug_title(bsc):
+    """
+    The bug description usually is after the "kernel live patch" message. Return a fixes message
+    if the livepatch name is not an ID from a bug on bugzilla.
+    """
+    if not bsc.isnumeric():
```
vmezzela, line 26: Perhaps we can change this after we merge #89 and take advantage of the `TEST_MODE` env? This would be a task for another PR though

vmezzela (approved):
Hi, it looks good to me! I left a comment just to discuss about a possible change that would be anyway a matter for another PR.

PR #89, opened 2025-02-19
Support GitHub CI

fgyanz:
Hi team,

This is the official PR integrating the testing into Github's CI/CD. From now on it will be possible to run `klp-build` and a subset of tests with the public [kernel-sources](https://github.com/openSUSE/kernel-source) automatically on Github. The job will be triggered for every push, so be mindful of the forced pushes and the likes.

Excluded stuff:
- No codestreams data. Meaning that the `setup` and `extract` will fail. For that reason all related tests are disabled.
- `klp-ccp`. Although it's possible to install it, there's currently no use for it (above reason).
- Anything related to OSC and IBS/OBS for obvious limitations in the environment.

Important considerations:
- The public kernel-source is mostly identical to the private one. There're pseudo-equivalent branches for any SLE version, with the sole exception of SLE12-SP5 and cve5.14/5.3. None of them are really critical for testing, and we can work with the above mentioned alternatives. All the rpm-* tags for each kernels are the same.
- We now have a  local configuration and `supported.csv` file, which gives a lot of flexibility during testing. We could now start working on writing more tests for the `scan` command.
- The kernel-source will be downloaded once every year (customizable, so let's see how it behaves) and then cached. Meaning that any push can reuse it without the need of running "git-clone".

For instance:
- Job once per year -> https://github.com/fgyanz/klp-build/actions/runs/13420110403/job/37490409263 (~20min)
- Normal job ->  https://github.com/fgyanz/klp-build/actions/runs/13420110403/job/37492062154 (~2min)


klpbuild/klplib/supported.py
```diff
@@ -41,6 +43,16 @@ def get_supported_codestreams():
...
 
+def __load_supported_file():
+    """
+    Read and return the lines of the supported file.
+
+    Returns:
+        list[str]: The list of lines of the supported file, excluding the
+        header.
+    """
+    with open("tests/supported.csv") as file:
```
marcosps, line 54: Does it work? I mean, using a relative open like this. If this works, then just ignore my comment :)

marcosps (approved):
You really made this kind of a big chance into very simple approach. Congratulations! LGTM!

I believe that now things will be much smoother, since we can easily spot new pylint and breaking tests with the CI integration.

klpbuild/klplib/supported.py
fgyanz: It does! Although it doesn't show in the CI jobs I shared, I did run a `scan` with a modern CVE, and worked great :)
In the Github's CI the structure is as follow:

klp-build/ (`$HOME)`:
- build/kernel-source
- klp/data
- klp/livepatches
- tests/
- and so on...

As klp-build repository is $HOME, it's fine using relative paths. In the configuration too.

vmezzela (approved):
Thanks @fgyanz , I like it! :rocket:

fgyanz:
@marcosps @vmezzela I did some minor changes in the PR.
1. The kernel-source repo was being only cached per PR, and not to the whole klp-build repo. I added a new YML that will trigger a job every time we push something to `devel`. A new kernel-source will be downloaded every month and then cached. This cached kernel-source should be accessible to all new PRs.
2. Since now the kernel-source will be up-to-date for all PRs, makes no sense to run the `fetch` command. So I disabled it during testing.
3. Unrelated, but now Pylint will error (and so the job too) if the score goes below 9.4 (currently at 9.46) and if there's any `Fatal` log message.

klpbuild/klplib/supported.py
marcosps: ok, thanks a lot!

marcosps (approved):
LGTM! Let's merge it and start opening more PRs to see how it works :)

Thanks a lot for working on it!

PR #92, opened 2025-02-25
TemplateGen enhancements

marcosps:
All these changes only add prototypes of _init/_cleanup, otherwise it's cleanups around TemplateGen class (which might die soon...)


vmezzela (approved):
Thanks @marcosps !

fgyanz (approved):
Very nice! LGTM!

PR #93, opened 2025-02-27
Module path resolving

marcosps:
I found this issue when started working on a new LP. And also I just added a new commit to remove extracted .zst files, as we do for xz.


fgyanz (approved):
LGTM! :)

PR #94, opened 2025-02-27
Convert format-patches to plugins

vmezzela:
Hi, nothing special here. Just continuing the work of converting commands to the new plugin system.

I noticed that this command takes the argument `lp_filter`, but never uses it. Should we just drop the option here?

While on it, do you have any other suggestion to improve this plugin?


marcosps:
> I noticed that this command takes the argument `lp_filter`, but never uses it. Should we just drop the option here?
I think so, yes.

vmezzela:
> > I noticed that this command takes the argument `lp_filter`, but never uses it. Should we just drop the option here?
>
> I think so, yes.

Done :heavy_check_mark:

marcosps (approved):
It works as expected. LGTM, thanks!

fgyanz (approved):
Tested and works! Thanks a lot :)
PS: Not something to improve on this PR, but perhaps having a template for the cover-letter would save some time.

PR #96, opened 2025-02-28
extract: Get gcc args by using compile_commands.json

marcosps:
Our make strategy worked well until we found the a new way of getting the gcc arguments used to compile a kernel source file. Recently it started to fail on newer codestreams, failing to find the proper _source_ directory from the -obj directory.

Use this opportunity to finally implement the arguments finding using compile_commands.json that is packaged by kernel-livepatch-default-livepatch-devel rpm.

The paths are extactly the same being found using make, but the instead of the absolute paths to the source directory of the codestreams, the compile_commands.json used '../'. Replacing ../ by the source directory of the affected codestream fixes the problem.

### DIFF
--- a/klpbuild/plugins/extractor.py
+++ b/klpbuild/plugins/extractor.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2021-2024 SUSE
 # Author: Marcos Paulo de Souza <mpdesouza@suse.com>
 
-import concurrent.futures
+from concurrent.futures import ThreadPoolExecutor
 import difflib as dl
 import json
 import logging
@@ -158,8 +158,7 @@ class Extractor():
             elif shutil.which("gcc-7"):
                 cc = "gcc-7"
             else:
-                logging.error("Only gcc12 or higher are available, and it's problematic with kernel sources")
-                raise
+                raise RuntimeError("Only gcc12 or higher are available, and it's problematic with kernel sources")
 
             make_args = [
                 "make",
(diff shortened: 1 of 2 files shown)


klpbuild/plugins/extractor.py
```diff
@@ -341,23 +341,25 @@ def apply_all_patches(self, cs, fil):
...
             if fname in d["file"]:
                 output = d["command"]
-                return Extractor.process_make_output(output)
+                # The arguments found on the file point to .., since they are generated
+                # when the kernel is compiled. Replace the .. by the codestream kernel source
+                # directory since klp-ccp needs to reach the files
+                return Extractor.process_make_output(output).replace("..", str(cs.get_src_dir()))
 
-        logging.error(f"Couldn't find cmdline for {fname}. Aborting")
+        logging.error("Couldn't find cmdline for %s. Aborting", fname)
```
vmezzela, line 362: Perhaps I'd rather not use here "Aborting", since if we don't find the command-line we're looking for in the json we just use the old method, right?
We could just emit a warning instead of an error, like:
```suggestion
        logging.warning("Couldn't find cmdline for %s in compile_commands.json, using get_make_cmd().", fname)
```
or something similar.

vmezzela (changes requested):

klpbuild/plugins/extractor.py
marcosps: We can do that, but we won't expect the make approach to work either, since compile_commands.json is created for each compiled source file. We can do what you suggested, but I won't expect to have more value from it.

vmezzela: I agree with you on this, my concern is just that as far as I understand we're not aborting here. Unless I'm missing something

vmezzela (approved):
lgtm :rocket:

fgyanz (approved):
Manually tested with CVE `2024-56648` (the one that triggered the bug) and works great! Thanks a lot!

PR #98, opened 2025-03-10
Fix paths read from compiler_commands.json

fgyanz:
Any path read from the `compile_commands.json` must be absolute, and if not converted to one. However, to make it absolute, only the first '..' should be replaced by the desired directory path. Any subsequent '..' must be ignored.

Bug triggered with CVE:
```
klp-build setup -n bsc1231196  --cve 2024-46815 --module amdgpu --conf
CONFIG_DRM_AMDGPU --file-func
drivers/gpu/drm/amd/display/dc/clk_mgr/dcn21/rn_clk_mgr.c rn_notify_wm_ranges
--arch x86_64
```

`klp-build extract -n bsc1231196 --apply-patches --filter '15.6rtu2'`

### DIFF
--- a/klpbuild/plugins/extractor.py
+++ b/klpbuild/plugins/extractor.py
@@ -347,10 +347,12 @@ class Extractor():
         for d in data:
             if fname in d["file"]:
                 output = d["command"]
-                # The arguments found on the file point to .., since they are generated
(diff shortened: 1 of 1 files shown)


marcosps (approved):
Thanks a lot for fixing it!

vmezzela (approved):
lgtm, thanks @fgyanz !

PR #99, opened 2025-03-20
Replace sources rpms with kernel.git

vmezzela:
This pull request modify `klp-build` so that the extractions relies on the `kernel.git` repository instead of the sources rpms.
Each source is no longer downloaded and extracted. Instead a worktree checked out to the respective git tag is created in place of the old source.

This means:
1. `klp-build` no longer downloads the sources rpm for **all** supported codestreams. In this way, setting up the tool for the first time is less time/disk-space consuming .
2. it only allocates a worktree when a specific source is needed for the extraction. The drawback is that when the extraction hits a codestream that doesn't have yet a worktree checked out, it'll take some time to create it.
3. to cleanup space, the command `klp-build clean-sources` will remove all the worktrees. This is useful when there are old trees that we no longer support.

Possible issues:
If for any reason when the worktree creation is interrupted, there might be problem of integrity with the kernel sources. It is recommended to run `klp-build clean-sources` and before the next extraction.

With this PR, a new entry in the config is required, `kernel_dir`, which contains the path to the git kernel repository (expanded tree).

To test the PR it's better to use a new `data_dir` to make sure we're not using old data.

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -7,7 +7,7 @@ from pathlib import Path, PurePath
 import re
 
 from klpbuild.klplib.utils import ARCH, get_workdir, is_mod, get_all_symbols_from_object, get_datadir
-from klpbuild.klplib.kernel_tree import init_cs_kernel_tree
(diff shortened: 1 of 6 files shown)


fgyanz (approved):
LGTM!

PR #100, opened 2025-03-21
ibs: Fix decompression of vmlinux and config for s390x

pmladek:
klp-build setup does not decompress vmlinux and config files from s390 rpms.

The problem is that cs.get_boot_file() returns x86_64 path by default. Fix it by explicitly passing the arch parameter.


fgyanz:
LGTM! Thanks a lot :)

PR #101, opened 2025-03-21
scan: Ignore code streams which do not match --filter

pmladek:
The packages for all code streams are currently downloaded and decompressed even when the caller is interested only into a particular one.

As a result, "klp-build setup" command runs much longer it could. It is not good in production and it makes debugging really painful.

The missing data are collected in scan(). Ignore the filtered code streams there.

IMPORTANT: I am not completely sure that it does not break some workflow. But the following
                        command succeeded with this change and was much faster:

klp-build -v setup --filter 15.3u42 --cve 2024-56600 --name bsc1235218 --conf CONFIG_IPV6 --file-func net/ipv6/af_inet6.c inet6_create

### DIFF
--- a/klpbuild/plugins/scan.py
+++ b/klpbuild/plugins/scan.py
@@ -4,6 +4,7 @@
 # Author: Marcos Paulo de Souza <mpdesouza@suse.com
 
 import logging
+import re
 import sys
 
 from klpbuild.klplib import utils
@@ -58,6 +59,12 @@ def scan(cve, conf, no_check, lp_filter, savedir=None):
         logging.info("Option --no-check was specified, checking all codestreams that are not filtered out...")
 
     for cs in all_codestreams:
+
+        # Skip filtered code streams
+        if lp_filter and not re.match(lp_filter, cs.name()):
+            logging.debug("	skipping code stream: %s", cs.name())
+            continue
+
         # Skip patched codestreams
         if not no_check:
             if cs.kernel in patched_kernels:


vmezzela:
Hi @pmladek, thanks for the PR. This indeed makes `klp-build` download only the needed rpms for a given cve, instead of all the codestreams we support. I just tested it :) 

I think we can apply it.

FWIW, by taking a look at this part of the code, I believe a good approach would need a rework of the IBS class, because the `lp_filter` is already passed to the constructor of `IBS`, before calling `IBS.download_cs_data`. See:

```py
    if data_missing:
        logging.info("Download the necessary data from the following codestreams:")
        logging.info("\t%s\n", " ".join(cs_missing))
        IBS("", lp_filter).download_cs_data(data_missing)
        logging.info("Done.")
```

Nevertheless, it seems `download_cs_data` is not using the filter at all. It's a bit confusing.. :/

Now, since it'd be also good to have a `download-data` command to download the data in advance and avoid the chicken and egg problem, I could try to rework this part while on it.

@fgyanz what do you think?

pmladek:
Honestly, I am not sure how the lp_filter was supposed to be used.
Maybe, it was introduced early and not maintained so that it stopped working at some point.
Or it was never really implemented.

Anyway, I like the idea to download everything in advance and avoid the chicken and egg problem.
And if the lp_filter stays working that it would be lovely ;-)

Should I do anything now?
Fernando also asked me to rebase this on top of the devel branch?
Should I still do it or is this going to be obsoleted, please?

fgyanz (approved):
Hi! The PR has green light from me :)
 @vmezzela makes sense. It could also be done in two steps, as adding the `download` command has higher priority over reworking existing code.

fgyanz:
Merged! Thanks @pmladek

PR #102, opened 2025-03-21
More info

pmladek:
I was running the script a newbie and I missed the following:

  + The connection between the names kernel-source and kgraft-patches branches.
      It might help a bit when we print also the name of the kernel source branch when
      listing the commits for the various code streams.

  + Also getting the information from the kernel-source branches takes a long time.
     It would give me a better feeling when it prints some progress with the -v option.


fgyanz (approved):
LGTM too! Tests passed. Thanks!

PR #103, opened 2025-03-26
Emit warning when optimization clones are found

vmezzela:
If the function to livepatch has an optimization clone, all its callers should be livepatched as well to avoid any of this clones from being left unpatched. Emit a warning when hitting this corner case.

We can merge this when the new version of `klp-ccp` using the ipa-clones gets released.

### DIFF
--- a/klpbuild/plugins/extractor.py
+++ b/klpbuild/plugins/extractor.py
@@ -447,16 +447,31 @@ class Extractor():
             cs.files[fname]["ibt"] = True
 
         out_log = Path(out_dir, "ccp.out.txt")
-        with open(out_log, "w") as f:
+        with open(out_log, "w+") as f:
             # Write the command line used
             f.write(f"Executing ccp on {odir}\n")
             self.print_env_vars(f, lenv)
             f.write("\n".join(args) + "\n")
             f.flush()
+
+            start_pos = f.tell()
             try:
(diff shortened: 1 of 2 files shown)


vmezzela:
Not sure the last update of the package contains this change though, I am running the up to date version and it's not analyzing the ipa-clones on the functions to patch yet.

Indeed, I pushed a test, but I couldn't test it because setting up a `tox` environment using the new version of `klp-ccp` it's a bit time consuming. We can just wait for it to be released and run it before merging.

fgyanz:
That's strange. I reproduced the ipa-clone's `null case` bug with the current klp-ccp package. It run the ipa-clone's script and everything.

vmezzela:
Yes, it makes sense: `klp-ccp` was already analyzing the `ipa-clones`, but it wasn't using the analysis yet since the list of patched functions was not yet supplied to the policies.
The `(null)` bug you triggered happens to be in the analysis part so it's triggered also by the old version of `klp-ccp`.

Instead, with the last update, it should use the result of the `ipa-clones` analysis to livepatch also the callers of inlined functions, although it still requires further improvement to fully automatize this.

As soon as the new version is released, we can run this test and merge the PR.

vmezzela:
@fgyanz we can merge now, the test is passing. I also reverted the first commit to check if the test fails as expected, and it does. So I think we're good.

I let you merge if you want to take another look at it.

fgyanz (approved):
LGTM!

PR #105, opened 2025-03-28
Add help message to some commands

vmezzela:


klpbuild/plugins/format_patches.py
```diff
@@ -15,7 +15,7 @@
 
 def register_argparser(subparser):
     fmt = subparser.add_parser(
-        "format-patches", help="SLE specific. Extract patches from kgraft-patches"
+        "format-patches", help="SLE specific. Format patches from kgraft-patches(_testscripts)?"
```
fgyanz, line 18: why the interrogation mark?

klpbuild/plugins/setup.py
```diff
@@ -18,7 +18,9 @@
 PLUGIN_CMD = "setup"
 
 def register_argparser(subparser):
-    setup = subparser.add_parser(PLUGIN_CMD)
+    setup = subparser.add_parser(
+        PLUGIN_CMD, help="SLE specific. Establish an initial working directory for a given livepatch"
```
fgyanz, line 22: I think we can remove the "SLE specific" from all the commands. It used to make sense when klp-build was not targeted only for suse kernels, but that's not the case anymore.

fgyanz (changes requested):

klpbuild/plugins/format_patches.py
vmezzela: Because the "_testscript" string is optional. Mmh perhaps I should do in some other way, like: `kgraft-patches[_testscripts]". What do you think?

fgyanz: Ah! I see. Yes, please. The [] sounds better.

fgyanz (approved):
LGTM!

Issue #24, opened 2024-09-30
Move away pro pkg_resources


marcosps:
This module is deprecated so we should move to its replacement.

marcosps:
Fixed by #79

PR #106, opened 2025-03-31
cmd: Add 'version' command

fgyanz:
Very simple PR that adds support for a version command. As the name suggest, it prints the current klp-build version.

```
$ klp-build -V
klp-build v0.0.3.post19+git.effd4d63
```

### DIFF
--- a/klp-build.1
+++ b/klp-build.1
@@ -9,6 +9,7 @@
 klp-build \- the kernel livepatching creation tool
 .SH SYNOPSIS
 .B klp-build
+[-h] [-v] [-V]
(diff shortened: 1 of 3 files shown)


vmezzela:
Hi @fgyanz , thanks for the PR. We really needed something like this!
Just one request, I remember in the past we were trying to avoid `pkg_resources` (see [here](https://github.com/SUSE/klp-build/issues/24)). Can't we use something else, like `importlib`, to retrieve the version?

fgyanz:
ah! good catch. Done :)

vmezzela (approved):
lgtm!

PR #107, opened 2025-04-03
format-patches: don't bail out if no testscript is found

vmezzela:


klpbuild/klplib/utils.py
```diff
@@ -346,7 +346,7 @@ def get_tests_path(lp_name):
         # as an entry point.
         return PurePath(test_dir_sh).parent
 
-    raise RuntimeError(f"Couldn't find {test_sh} or {test_dir_sh}")
```
fgyanz, line 349: would it be possible to keep this as a warning? And perhaps rephrase it so that it doesn't sound like an error. Just for completeness.

fgyanz (changes requested):

klpbuild/klplib/utils.py
vmezzela: Sure, we can also have it as `logging.debug` so it shows up only when using `--verbose`.
What do you think is better?

fgyanz: humm  I'd vote for a `warn`. Probably this message will be most useful to advanced users (mainly us). Like, "hey, Careful there, your are not sending a test". And we don't usually run `format-patches` with `--verbose`.

But not a big deal anyway. Just my 2cents :)

fgyanz: BTW, offtopic, but:
`klp-build -v format-patches -n xxxxx -v 1`

With verbose enabled we have two `v`s. Can be a bit misleading.

vmezzela: > humm I'd vote for a warn

Okay for me, let me update the PR :) 

> With verbose enabled we have two vs. Can be a bit misleading.

Oh true, but thankfully the `verbose` here is not really needed.

vmezzela: Done :heavy_check_mark:

fgyanz: thanks!

fgyanz (approved):
LGTM!

PR #109, opened 2025-04-03
Drop inline plugin

vmezzela:


fgyanz (approved):
LGTM! Thanks @vmezzela :)

PR #110, opened 2025-04-07
Automate patch analysis phase

fgyanz:
Hi team,

This PR automates most of the manual patch analysis done between the `scan` and `setup` commands. It gathers a lot of useful information, saving some time during development.

Full list of features:
- Patch analysis. Retrieve all the files modified by each of the backported patches. Each of those files are also analyzed, gathering the corresponding CONFIG and kernel module.
- Codestream grouping by files, configs and modules.
- klp-build can now check if a module is supported in a specific kernel version (as per the `supported.conf`).
- Better correlation between CONFIGs and modules, per architecture.

Things to be taken into consideration:
 - As a first version, there are probably many corner cases I'm not aware of. 
 - No automated testing yet. As it is already quite a big PR, I'll try to have them on a different one.
 - ~~Relies on the expanded kernel tree or extracted rpm sources.  This can be improved so that it uses the kernel.git repo instead.~~

Example output for a CVE affecting an unsupported kernel module:
```
$ klp-build scan --cve 2024-49931
Downloading codestreams file
Fetching changes from all supported branches...
Getting SUSE fixes for upstream commits per CVE branch. It can take some time...

12.5: SLE12-SP5
None

15.3: SLE15-SP3-LTSS
None

15.4: SLE15-SP4-LTSS
None

15.5: SLE15-SP5-LTSS
None

15.6: SLE15-SP6
e419da4941b0c5e1b4d545528180bfa2fd0341fa

15.6rt: SLE15-SP6-RT
e419da4941b0c5e1b4d545528180bfa2fd0341fa

6.0: SUSE-2024
e419da4941b0c5e1b4d545528180bfa2fd0341fa

6.0rt: SUSE-2024-RT
e419da4941b0c5e1b4d545528180bfa2fd0341fa

cve-5.3: cve/linux-5.3-LTSS
None

cve-5.14: cve/linux-5.14-LTSS
None

upstream
e106b7ad13c1 ("wifi: ath12k: fix array out-of-bound access in SoC stats")

Searching for already patched codestreams...
Checking filter...
Initiating patch analysis...

[*] Analysing modified files...

6.0rtu2-3 6.0u2-3 15.6rtu0-4 15.6u0-5:
FILE: drivers/net/wireless/ath/ath12k/dp_rx.c
CONF: CONFIG_ATH12K
OBJ: drivers/net/wireless/ath/ath12k/ath12k

[*] Analysing required CONFIGs...

6.0rtu2-3 6.0u2-3 15.6rtu0-4:
CONF: CONFIG_ATH12K
x86_64: m
ppc64le: n
s390x: n

15.6u0-5:
CONF: CONFIG_ATH12K
x86_64: m
ppc64le: m
s390x: n

[*] Analysing affected kernel modules...

6.0rtu2-3 6.0u2-3 15.6rtu0-4 15.6u0-5:
MOD: drivers/net/wireless/ath/ath12k/ath12k
Supported: False

Skipping codestreams with unsupported kernel modules:
	6.0rtu2-3 6.0u2-3 15.6rtu0-4 15.6u0-5
	6.0rtu4 6.0u4 15.6rtu5-9 15.6u6-9
Skipping unaffected codestreams (missing backports):
	12.5u54-66 15.3u42-55 15.4u24-38 15.5u11-24
All supported codestreams are already patched. Exiting klp-build
```

### DIFF
--- a/klpbuild/klplib/file2config.py
+++ b/klpbuild/klplib/file2config.py
@@ -32,25 +32,27 @@ def _sanitize_config(target):
     return config
 
 
-def _find_config(cs, obj_path, deep):
-
+def _find_config(cs, base_dir, relative_obj_path, deep):
     if deep > 10:
         return None, ""
 
-    make_file = Path(obj_path.parent, "Makefile")
+    make_file = Path(base_dir, "Makefile")
 
     lines = _load_makefile(cs, make_file)
 
-    obj_name = PurePath(obj_path).name
+    if not lines and Path(".") != base_dir:
+        relative_obj_path = base_dir.name + "/" + relative_obj_path
+        return _find_config(cs, base_dir.parent, relative_obj_path, deep+1)
+
     for line in lines:
         sep = line.split()
-        if obj_name not in sep:
+        if relative_obj_path not in sep:
             continue
 
         # target found, check if this one with config
         target = sep[0]
         if target.startswith('obj-'):
-            return _sanitize_config(target), str(obj_path.with_suffix(''))
+            return _sanitize_config(target), str((base_dir/relative_obj_path).with_suffix(''))
 
         # target contains another object file rule, so strip it would and try
         # again
@@ -60,7 +62,7 @@ def _find_config(cs, obj_path, deep):
             # print(ve)
             continue
 
-        return _find_config(cs, Path(obj_path.parent, target + '.o'), deep + 1)
+        return _find_config(cs, base_dir, target + '.o', deep + 1)
 
     return None, ""
 
@@ -76,8 +78,8 @@ def find_configs_for_files(cs, file_paths: list):
 
     for path in file_paths:
         path = path.strip()
-        obj_file = path.replace('.c', '.o')
-        config, obj = _find_config(cs, Path(obj_file), 0)
+        obj_file = Path(path.replace('.c', '.o'))
+        config, obj = _find_config(cs, obj_file.parent, obj_file.name, 0)
         if not config:
             missing.append(path)
         elif config == 'y' or config == 'm':
--- a/klpbuild/klplib/file2config.py
+++ b/klpbuild/klplib/file2config.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2024 SUSE
+# Author: Fernando Gonzalez <fernando.gonzalez@suse.com>
+#
+# Copied from kernel-source:scripts/file2config.py
+
+import re
+
+from pathlib import Path, PurePath
+from os import chdir
+from argparse import ArgumentParser
+from os.path import expanduser
+
+def _load_makefile(cs, make_file: str) -> list:
+
+    if not cs.check_file_exists(make_file):
+        return []
+
+    buffer = cs.read_file(make_file)
+    assert buffer
+
+    joined = re.sub(r'\\\s*\n[^:]', ' ', buffer)
+
+    lines = joined.split('\n')
+
+    return lines
+
+
+def _sanitize_config(target):
+    config = target.strip('+=').strip().strip('obj-$(){}:').strip()
+    return config
+
+
+def _find_config(cs, obj_path, deep):
+
+    if deep > 10:
+        return None, ""
+
+    make_file = Path(obj_path.parent, "Makefile")
+
+    lines = _load_makefile(cs, make_file)
+
+    obj_name = PurePath(obj_path).name
+    for line in lines:
+        sep = line.split()
+        if obj_name not in sep:
+            continue
+
+        # target found, check if this one with config
+        target = sep[0]
+        if target.startswith('obj-'):
+            return _sanitize_config(target), str(obj_path.with_suffix(''))
+
+        # target contains another object file rule, so strip it would and try
+        # again
+        try:
+            target, _ = target.rsplit('-', 1)
+        except ValueError as ve:
+            # print(ve)
+            continue
+
+        return _find_config(cs, Path(obj_path.parent, target + '.o'), deep + 1)
+
+    return None, ""
+
+
+def find_configs_for_files(cs, file_paths: list):
+
+    configs = dict()
+    build_ins = []
+    missing = []
+
+    if not file_paths:
+        return configs, build_ins, missing
+
+    for path in file_paths:
+        path = path.strip()
+        obj_file = path.replace('.c', '.o')
+        config, obj = _find_config(cs, Path(obj_file), 0)
+        if not config:
+            missing.append(path)
+        elif config == 'y' or config == 'm':
+            build_ins.append(path)
+        elif config.startswith('CONFIG_'):
+            configs[path] = {'config': config, 'obj': obj}
+        # else there is garbage like 'subst', 'vds' for wrongly parsed input
+
+    return configs, build_ins, missing
+
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -475,3 +475,45 @@ def ksrc_read_file(kernel_version, file_path):
                           f"{ksrc_tag}:{file_path}"],
                          capture_output=True, text=True)
     return ret.stdout
+
+
+def ksrc_is_module_supported(module, kernel):
+        """
+        Check if a kernel module is supported on a specific kernel.
+        This is done by reading the 'supported.conf' file.
+
+        Args:
+            module (str): Full path of the module.
+            kernel (sr): Kernel version.
+
+        returns:
+            Return True if supported. False otherwise.
+            """
+
+        mpath = module
+        prev = ""
+        idx = 1
+        supported = False
+
+        out = ksrc_read_file(kernel, "supported.conf").splitlines()
+        if not out:
+            return False
+
+        # Try the following path combinations to see if it matches with
+        # any rule in the supported.conf:
+        #   my/kernel/module/path
+        #   my/kernel/module/*
+        #   my/kernel/*
+        #   my/*
+        while mpath != prev:
+            r = re.compile(rf"^[-+\s].*{mpath}")
+            match = list(filter(r.match, out))
+            if match:
+                supported = match[0][0] != '-'
+                break
+
+            prev = mpath
+            mpath = module.rsplit("/", idx)[0] + r"/\*"
+            idx += 1
+
+        return supported
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -493,7 +493,7 @@ def ksrc_is_module_supported(module, kernel):
         mpath = module
         prev = ""
         idx = 1
-        supported = False
+        supported = True
 
         out = ksrc_read_file(kernel, "supported.conf").splitlines()
         if not out:
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -100,14 +100,19 @@ def diff_commits(base, new, patch, pattern=""):
 
     return diff.returncode
 
-def get_commit_files(commit, regex=None):
+
+def get_commit_files(commit, inside_patch=False, regex=r"patches\.suse\/.+\.patch"):
     """
-    Get the files that have been modified in one specific commit.
+    Get the files that have been modified in one specific commit or
+    within the patch files of the commit.
     Optionally only get those that match the given regular expression.
 
     Args:
         commit (str): The commit to be anylized.
         regex (str): Optional regex.
+        inside_path (bool): True for getting the files modified by the
+        patch file in the commit. False for just getting the files in the
+        commit.
 
     returns:
         List: Return the files that match the regex, if set. Otherwise,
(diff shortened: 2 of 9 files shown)


klpbuild/plugins/scan.py
```diff
@@ -88,9 +108,12 @@ def scan(cve, conf, no_check, lp_filter, download, savedir=None):
                 tmp_working_cs.append(cs)
         working_cs = tmp_working_cs
 
+    if unsupported:
+        logging.info(f"Skipping codestreams with unsupported kernel modules:")
+        logging.info("\t%s", utils.classify_codestreams_str(unsupported))
 
     if conf_not_set:
-        logging.info("Skipping codestreams without %s set:", conf)
+        logging.info(f"Skipping codestreams without {conf} set:")
```
vmezzela, line 116: Just a nit, can we use the `%s` in logging? I remember the linter complaining about using this format. I suspect this hunk is due to a rebase from old version of the code, perhaps you can just drop it.

vmezzela (commented):
Hi Fernando, I've done an initial review iteration, it's a great improvement to `klp-build`! Thanks for it!

I'll use this branch for the next LP to test it further :)

klpbuild/klplib/patch.py
```diff
@@ -0,0 +1,199 @@
...
+        cs_list (list): List of affected codestreams.
+        sle_commits (dict): List of commits by codestream.
+        gh: Github helper object.
+    '''
+
+    report = {}
+
+    for cs in cs_list:
+        srcdir = cs.get_src_dir()
+        if not os.path.isdir(srcdir):
```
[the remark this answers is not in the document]
fgyanz: FWIW this was intentional. The idea was to gently provide as much info of the patches as possible, but keeping the output clean, without errors or warnings. The `scan` command should go as smooth as possible, I believe. In case the patch analysis fails to find useful info, we fallback to report the codestream as affected and leave the work to the user.

However, It was a bad idea to rely on the expanded trees. I've changed that part of the code to, instead of reading the files from the expanded kernel tree, do it directly from the kernel.git. This avoids expanding the trees, and can also be used in combination with klp-bugzilla.

I've done some measurements and speed-wise is on par to the best-case scenario of the previous version.

klpbuild/klplib/patch.py
```diff
@@ -0,0 +1,199 @@
...
+
+            for file in missing:
+                key = f"{file}::"
+                if not report.get(key) or cs not in report[key]:
+                    report.setdefault(key, []).append(cs)
+
+            for file, dat in fconfigs.items():
+                key = f"{file}:{dat['config']}:{dat['obj']}"
+                if not report.get(key) or cs not in report[key]:
+                    report.setdefault(key, []).append(cs)
```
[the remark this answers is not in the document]
fgyanz: Good idea, thanks! I changed it to use `defaultdict(list)` as suggested. The `set` idea sounds great too, but perhaps making it work is out of the scope of this PR.

klpbuild/plugins/scan.py
fgyanz: Ah! my bad, fixed. Thanks!

marcosps: I believe that it's the otherway around: it's suggested to not use the `f` formatting, and use the `%` instead.

fgyanz:
Hi @vmezzela , thanks a lot for the review! I did some changes based on your comments. Also, feel free to say if you would want the code organized in a different way, or a better file/function naming.

klpbuild/plugins/scan.py
fgyanz: actually, I misread @vmezzela 's comment and did it as marcos's said. Hum not sure what to do now :D

klpbuild/plugins/scan.py
```diff
@@ -77,8 +80,25 @@ def scan(cve, conf, no_check, lp_filter, download, savedir=None):
     if conf or download:
         download_missing_cs_data(working_cs)
 
+    # Automated patch analysis phase. Not compatible with --conf.
+    if not conf and commits:
```
marcosps, line 84: Maybe invert the 

`if commits and not conf:`

? At least when mind reading it seems that both need to False, at least for me :P of course not a blocker for the patches, just a small nit

klpbuild/plugins/scan.py
```diff
@@ -88,9 +108,12 @@ def scan(cve, conf, no_check, lp_filter, download, savedir=None):
                 tmp_working_cs.append(cs)
         working_cs = tmp_working_cs
 
+    if unsupported:
+        logging.info(f"Skipping codestreams with unsupported kernel modules:")
+        logging.info("\t%s", utils.classify_codestreams_str(unsupported))
 
     if conf_not_set:
-        logging.info("Skipping codestreams without %s set:", conf)
+        logging.info(f"Skipping codestreams without %s set:", conf)
```
marcosps, line 116: The `f` was added by mistake here?

klpbuild/plugins/scan.py
vmezzela: yep, I meant that: not using `f` but `%s` instead, so the same as Marcos's saying :)

klpbuild/klplib/patch.py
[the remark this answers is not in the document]
vmezzela: I also prefer this approach, I would expand the tree only and only if we need to extract the livepatch :)

klpbuild/plugins/scan.py
fgyanz: Makes sense. Done!

klpbuild/klplib/patch.py
```diff
@@ -0,0 +1,196 @@
...
+    by the backported patches.
+    For each of the files it retrieves the corresponding kernel module
+    and config.
+    Lastly, it returns back a report grouping the codestreams by files,
+    modules and configs.
+
+    Args:
+        cs_list (list): List of affected codestreams.
+        sle_commits (dict): List of commits by codestream.
+        gh: Github helper object.
```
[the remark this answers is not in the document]
fgyanz: Oops sorry! Corrected.

klpbuild/plugins/scan.py
fgyanz: Indeed... I was in a hurry and forgot to remove that. Thanks! and Done :)

vmezzela:
One thing that just came to my mind, we should move `update_kernel_tree_tags` from `setup` to `scan`. Right? @fgyanz

fgyanz:
Agreed. Done on #112 :+1:

klpbuild/plugins/scan.py
```diff
@@ -77,8 +80,25 @@ def scan(cve, conf, no_check, lp_filter, download, savedir=None):
...
+
+        logging.info("[*] Analysing required CONFIGs...\n")
+        configs_report = patch.analyse_configs(working_cs)
+        patch.print_configs(configs_report)
+        conf_not_set, conf = patch.filter_unset_configs(working_cs)
+
+        logging.info("[*] Analysing affected kernel modules...\n")
+        kmodules_report = patch.analyse_kmodules(working_cs, gh)
+        patch.print_kmodules(kmodules_report)
+        unsupported = patch.filter_unsupported_kmodules(working_cs)
```
vmezzela, line 98: Hi, part of this relies on having downloaded the rpms beforehand  I guess. Shouldn't we call `download_missing_cs_data` before this code block get executed?

Though I believe if we can make this working without the rps would be even better.

klpbuild/klplib/patch.py
```diff
@@ -0,0 +1,198 @@
...
+
+    report = defaultdict(list)
+
+    for cs in cs_list:
+        for _, dat in cs.files.items():
+            conf = dat['config']
+            if conf in cs.configs:
+                continue
+
+            cs.configs[conf] = cs.get_all_configs(conf)
```
vmezzela, line 88: `get_all_configs` here needs the data. Can we modify this to read `kernel-source/config/x86_64/default` from the respective rpm tag?

vmezzela (changes requested):
I left two comments inline :)

klpbuild/klplib/patch.py
vmezzela: I'm not 100% sure there's a 1:1 relation betweeen the config from the rpms and this one, we better check it :)

klpbuild/klplib/file2config.py
```diff
@@ -0,0 +1,90 @@
...
+
+    return lines
+
+
+def _sanitize_config(target):
+    config = target.strip('+=').strip().strip('obj-$(){}:').strip()
+    return config
+
+
+def _find_config(cs, obj_path, deep):
```
vmezzela, line 35: This do not work in case the Makefile is not present in the same directory of the file. I updated the `file2config.py` script in `kernel-source` to fix that (see `f4013908416f82efbcad5c15cc6c8f21a47d41eb`) and I tried to backport the fix here as well, if the patch is okay for you please feel free to pull it from [here](https://github.com/vmezzela/klp-build/tree/fix-file2config) and reorganize as you prefer. :smiley: 

`klp-build scan --cve 2024-58014` is a testcase

vmezzela (changes requested):
Do you mind dropping the `exec` bit in `file2config.py`?

Thanks! :)

klpbuild/klplib/file2config.py
fgyanz: Thanks a lot for improving the script! I cherry-picked your commit and tested it. Worked correctly as expected :)

klpbuild/klplib/patch.py
fgyanz: I agree. However, I would propose having those changes on a different PR.  This one is already too big and seems to be working fine. I'd rather not add unpredictable variables without further testing. Note that fetching the config from kernel-source will be needed for CI testing, so perhaps we can kill two birds with one stone by having a single PR for testing and "kernel-source configs".

klpbuild/plugins/scan.py
fgyanz: As it is, it will download only when the `--download`  is specified. Otherwise, there would be no reason to have such option.

fgyanz:
Thanks for the comments! I pushed some minor changes fixing bugs and corner cases.

klpbuild/plugins/scan.py
vmezzela: I agree with you that `scan` should not download anything. 

But with the PR as is  `scan` fails if the data are not present, because get_all_configs will raise an error (`FileNotFound` IIRC).

You can try to change the data_dir in the config to an empty directory to check the failure.

vmezzela (changes requested):

klpbuild/plugins/scan.py
fgyanz: yes, that is true. But note that for avrg users, the recommended `scan` command as per the documentation is: `klp-build scan --download`. As you know, it will download the rpms on a per need basis and run the patch analysis. 

For automated environments such as `klp-bugzilla`, I'd rather not risk downloading stuff while 100s of processes are accessing the `data/` dir (IIRC we have some locking mechanism, but who knows if it works).
The next best option is crashing, at least we won't get incorrect reports from klp-build. However, to somewhat avoid that, there's the `klp-build data --download`.

I know this is far from a good solution... This is just a temporal workaround until we implement reading the configs from the kernel-source :)

vmezzela: Mmmh, merging as is means that we'd have a broken scan, right?

For instance, the other day I've run a scan (using this branch) and it failed because i didn't  have the data for the latest update of MICRO and the only way to `scan` was was to download them (with the `--download` option). So, there would be no way to run the `scan` without necessarily downloading the data.

I would be more incline to fix this issue before merging the PR since it is related. But I am okay, though I would implement reading the configs from the kernel-source before we merge devil into main otherwise we'll have a scan command that fails and the only way to make it work if you need to run `scan` is to download the data.

fgyanz: The PR still has a long way to go before merging IMHO. I'm in no rush at all, that's way I also proposed on some previous comment to have a different PR for the config and testing. After all, we cannot test this PR properly without reading the config from ksrc. So we kill two birds with one stone :)

vmezzela: Okay then I totally agree with you, sorry for the misunderstanding :)

fgyanz: No problem, glad to know we came to an agreement! I'll mark this as resolved, but in the meantime someone should work on the config-ksrc.

marcosps (approved):
LGTM!

klpbuild/klplib/ksrc.py
```diff
@@ -450,3 +469,45 @@ def ksrc_read_file(kernel_version, file_path):
                           f"{ksrc_tag}:{file_path}"],
                          capture_output=True, text=True)
     return ret.stdout
+
+
+def is_module_supported(module, kernel):
```
marcosps, line 474: Not a big deal, but, wouldn't be better is we passed a codestream object and not the kernels string directly? I mean, to create the tests for this function the string is much easier to do, but we don't pass kernels trings anywhere else on klp-build.

@fgyanz what do you think?

vmezzela: If we want to pass the codestream object, then I would do it in the codestream class.
Like:
`cs.is_module_supported(module)` -> `is_module_supported(module, kernel)` (this one from the ksrc module)
leaving the method in ksrc unchanged, but just passing though

fgyanz: Hey guys! I like the idea. I'd leave the one in ksrc as is for testing purposes and for the reason Vincenzo already explained. However, for `get_commit_files` I think it doesn't make much sense to have a codestream wrapper.

fgyanz: Done! `cs.is_module_supported() `-> `ksrc_is_module_supported()`

klpbuild/klplib/patch.py
```diff
@@ -0,0 +1,200 @@
...
+    '''
+    Function that analyses, per codestream, each of the kernel modules found
+    in analyse_files().
+    For each of the modules it checks if they are supported, builtin and so on.
+    Lastly, it returns back a report grouping the codestreams by modules and
+    their configuration.
+
+    Args:
+        cs_list (list): List of affected codestreams.
+        gh: GitHelper object.
```
marcosps, line 147: The gh argument was removed, so I believe this is a leftover.

marcosps (approved):
Besides the nit, LGTM!

vmezzela (approved):
lgtm!

PR #113, opened 2025-04-14
klplib: utils: don't use PurePath in get_tests_path

vmezzela:
PurePath doesn't have is_file(), so prepare-test fails when the test script is contained in a directory


fgyanz (approved):
LGTM! Thanks!

PR #114, opened 2025-04-14
scan: Filter codestreams with no-check enabled

fgyanz:
Scan was ignoring the filtered codestreams when running with 'no-check' enabled.
The t`est_valite_conf_unsupported_arch` was failing because it expected a filtered codestream-related output.


marcosps (approved):
Thanks for the fix, LGTM!

vmezzela (approved):
lgtm! thanks!

PR #115, opened 2025-04-14
setup.py: Remove importlib as dependency

marcosps:
importlib is part of python3, and leaving this depency here makes it fail to be installed on openSUSE Tumbleweed.


fgyanz (approved):
LGTM!  On Leap I was also having the same issue. Thanks!

vmezzela (approved):
lgtm! thanks!

PR #117, opened 2025-04-16
klplib: kernel_tree: use decorator to update kernel tree tags

vmezzela:
So we don't have to remember every time we use the kernel tree in a new part of the codebase :)


marcosps (approved):
I'm amazed how the decorators work. Thanks a lot!

fgyanz (approved):
LGTM! Thanks! :)

PR #120, opened 2025-04-23
klplib: codestream: use ksrc to read config

vmezzela:
This PR is a prerequisite for #110 .

Please @fgyanz , notice that I have copied the `read_file_in_tag` function from you PR and moved in `klplib/utils` so that it can be used both from `ksrc` and `kernel_tree`.

### DIFF
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -441,3 +441,12 @@ def cs_is_affected(cs, cve, commits):
 
     return len(commits[cs.name_cs()]["commits"]) > 0
 
+
+def ksrc_read_file(kernel_version, file_path):
+    ksrc_dir = get_user_path("kernel_src_dir")
+    ksrc_tag = "rpm-" + kernel_version
+
+    ret = subprocess.run(["git", "-C", ksrc_dir, "show",
+                          f"{ksrc_tag}:{file_path}"],
+                         capture_output=True, text=True)
+    return ret.stdout
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -6,6 +6,7 @@
 from pathlib import Path, PurePath
 import re
 
+from klpbuild.klplib.ksrc import ksrc_read_file
 from klpbuild.klplib.utils import ARCH, get_workdir, is_mod, get_all_symbols_from_object, get_datadir
 from klpbuild.klplib.kernel_tree import init_cs_kernel_tree, file_exists_in_tag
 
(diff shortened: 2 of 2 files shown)


klpbuild/klplib/ksrc.py
```diff
@@ -441,3 +441,9 @@ def cs_is_affected(cs, cve, commits):
 
     return len(commits[cs.name_cs()]["commits"]) > 0
 
+
+def ksrc_read_file(kernel_version, file_path):
+    kernel_tree = get_user_path("kernel_src_dir")
```
fgyanz, line 446: Shouldn't the variable be named `kern_src` ?

fgyanz: same in the line below. Perhaps something like `tag`.

klpbuild/klplib/utils.py
```diff
@@ -351,3 +352,10 @@ def get_tests_path(lp_name):
 
 def in_test_mode():
     return os.getenv("TEST_MODE", 'n') == 'y'
+
+
+def read_file_in_tag(git_directory, branch, file_path):
+    ret = subprocess.run(["git", "-C", git_directory, "show",
+                          f"{branch}:{file_path}"],
+                         capture_output=True, text=True)
+    return ret.stdout
```
fgyanz, line 361: I'm not really fond of having this function in `utils`. As I see it, `utils` is a last resource, and we should avoid moving functions here as much as possible.
Also, we don't really save that much space by decoupling  `read_file_in_tag` from the main functions. I'd rather have this 3 lines repeated than making `utils` beast grow in size :) 
What do you think?

fgyanz (changes requested):
Except for a few details the PR LGTM :) Thanks a lot! @vmezzela 
For the record, do you mind pointing out the differences, if any, between the configs in kernel-source and in rpms?

klpbuild/klplib/ksrc.py
marcosps: agreed

marcosps (approved):
I don't see any other problem besides the variable name change!

klpbuild/klplib/utils.py
fgyanz: I know I'm nitpicking here. But since we are doing so much cleaning and reworking, I feel obligated to point out the issue with `utils`. It's generally known to be a bad programing practice to rely too much on `utils`, and should mostly be kept for parsing functions and related stuff (as we currently mostly do). I do agree that having  `read_file_in_tag` accessible to any file is a good idea. My concern is having it in `utils` :)

That being said, I'm not willing to die on this hill ^^ I'll conform with the majority.

marcosps: Hum... didn't know about this bad practice, mas as I said, either way I'm happy :)

klpbuild/klplib/ksrc.py
vmezzela: Oh, my bad. Wrong copy and paste :smile: 

I'll fix it.

klpbuild/klplib/utils.py
vmezzela: Hi, I agree that it was not a good idea to move it to `utils`, I'll move it back to ksrc :)

vmezzela:
> Except for a few details the PR LGTM :) Thanks a lot! @vmezzela For the record, do you mind pointing out the differences, if any, between the configs in kernel-source and in rpms?

Oh sure, I forgot! that's an important point! I should mention it also in the changelog of the commit.



> I don't see any other problem besides the variable name change!

Thank guys for the review!

fgyanz (approved):
Thanks! LGTM! :)

PR #121, opened 2025-04-24
Ibs cleanup

marcosps:
guys, the PR is big, but the interesting part is on the last patch :)

I'll test it more tomorrow, but I believe that it's working :)

### DIFF
--- a/klpbuild/klplib/ibs.py
+++ b/klpbuild/klplib/ibs.py
@@ -48,6 +48,15 @@ def do_work(func, args):
                 logging.error(result)
 
 
+# We can try delete a project that was removed, so don't bother with errors
+def delete_built_rpms(cs, lp_name):
+    try:
+        for arch in cs.archs:
+            shutil.rmtree(Path(cs.get_ccp_dir(lp_name), arch, "rpm"), ignore_errors=True)
+    except KeyError:
+        pass
+
+
 class IBS():
     def __init__(self, lp_name, lp_filter):
         self.osc = Osc(url="https://api.suse.de")
@@ -368,14 +377,6 @@ class IBS():
 
             logging.info("Done.")
 
-    # We can try delete a project that was removed, so don't bother with errors
-    def delete_rpms(self, cs):
-        try:
-            for arch in cs.archs:
-                shutil.rmtree(Path(cs.get_ccp_dir(self.lp_name), arch, "rpm"), ignore_errors=True)
-        except KeyError:
-            pass
-
     def download_built_rpms(self, lp_name):
         rpms = []
         i = 1
@@ -391,7 +392,7 @@ class IBS():
                 continue
 
             # Remove previously downloaded rpms
-            self.delete_rpms(cs)
+            delete_built_rpms(cs, lp_name)
 
             archs = result.xpath("repository/arch")
             for arch in archs:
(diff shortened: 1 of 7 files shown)


klpbuild/klplib/ibs.py
```diff
@@ -28,606 +28,627 @@
...
-        self.osc = Osc(url="https://api.suse.de")
+def convert_prj_to_cs(prj, prefix):
+    return prj.replace(f"{prefix}-", "").replace("_", ".")
 
-        self.lp_name = lp_name
-        self.lp_filter = lp_filter
 
-        self.ibs_user = self.osc.username
-        self.prj_prefix = f"home:{self.ibs_user}:{self.lp_name}-klp"
+def cs_to_project(cs, prefix):
```
vmezzela, line 35: Do you mind changing it to `convert_cs_to_proj` so it's consistent with `convert_proj_to_cs`?

klpbuild/klplib/ibs.py
```diff
@@ -28,606 +28,627 @@
...
-            # Do not push .osc directory
-            if ".osc" in str(fname):
+    osc = Osc(url="https://api.suse.de")
+    prefix = prj_prefix(lp_name, osc)
+
+    logging.info(osc.build.get_log(cs_to_project(cs_list[0], prefix), "standard", arch, "klp"))
+
+
+def status(lp_name, lp_filter, wait=False):
+    finished_prj = []
```
vmezzela, line 466: Can we move this function to `plugins/status.py`? What do you think?

vmezzela (commented):
Hi Marcos, thanks for this PR! I did an initial review and left some comments inline but are just nitpicks.

I will use this PR for the next LP the test it further as reviewing it from the changes it's quite difficult :smile:

klpbuild/klplib/ibs.py
marcosps: No, I'll do it

klpbuild/klplib/ibs.py
marcosps: I though about it, but the push command also uses it... Maybe I'll move the status command here and then make push command import from status.py

klpbuild/klplib/ibs.py
vmezzela: Thanks :)

klpbuild/klplib/ibs.py
vmezzela: oh, didn't notice it's used by `push` as well. Then I think the best option would be to move all the ibs commands to their related plugin module and make `push` call `status`, because I'd rather not call a plugin from `klplib`.

But I think this can wait another PR

klpbuild/plugins/extractor.py
```diff
@@ -382,6 +382,7 @@ def cmd_args(self, cs, fname, out_dir, fdata, cmd):
             "-mindirect-branch-cs-prefix",
             "-mharden-sls=all",
             "-fmin-function-alignment=16",
+            "-Wno-dangling-pointer",
```
vmezzela, line 385: Can we remove this hunk from here and create a new commit for it? I'd apply it to `main` as well

fgyanz (approved):
LGTM! Thanks :)

PR #122, opened 2025-04-25
klplib/ksrc: Filter out merge commits

fgyanz:
In some cases git-log incorrectly reports back the commit that merged the branch containing the expected commit, and not the expected commit itself. This happens even with the "--no-merge" flag enabled, so there's no choice but to skip them manually.

Also, in order to find the commit that introduced the patch, from now on git-log will be forced to follow all the parent commits and traverse the full patch's history, until file creation. This is a more exhaustive method, and as such slower than before (~4 second slowdown).

I'd like to point out that the ideal git-log flag would be `--follow`, but unfortunately it's just too slow for our propose  (~1:32min). Hence, the second best option, `full-history`.

### DIFF
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -232,6 +232,8 @@ def get_commits(cve, savedir=None):
                         "-C",
                         kern_src,
                         "log",
+                        "--full-history",
+                        "--remove-empty",
                         "--numstat",
                         "--reverse",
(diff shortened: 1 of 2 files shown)


marcosps (approved):
Amazing that we are still finding corner cases... thanks a lot for writing a test for it!

vmezzela (approved):
lgtm! thanks :rocket:

PR #124, opened 2025-05-05
-Wno-dangling-pointer removal

marcosps:
Again klp-ccp is too slow to keep up with the gcc flags used by "modern" kernels, so we need to do something on our side, otherwise we can't produce livepatches.


vmezzela (approved):
lgtm!

fgyanz (approved):
lgtm! Thanks! :)

PR #125, opened 2025-05-14
klplib: ibs: Fix download_and_extract new tuple element

marcosps:
This is needed otherwise it will fail whenever we need to download any rpms on setup phase.


fgyanz (approved):
LGTM!

vmezzela (approved):
lgtm!

PR #126, opened 2025-05-27
extractor.py: Ignore 'make' errors

fgyanz:
Ignore errors as long as we manage to get back the correct make commands for the target object.
This PR is a workaround to allow klp-build work on older distributions where "make" was failing.


marcosps (approved):
LGTM

vmezzela (approved):

PR #128, opened 2025-05-29
plugins: extractor: match more compiler commands

vmezzela:
My concern is: is that okay that something like `gcc-13` ends up there? Or do we have an issue somewhere else?

I wouldn't expect to see that version of gcc.

(this should be backported also to `main`)


marcosps (approved):
LGTM, the other approach was made to all current codestreams, but it changed, so makes sense to remove the gcc command using a regex now

fgyanz (approved):
LGTM!

PR #131, opened 2025-06-27
Convert extract into a proper plugin

marcosps:
Now we are covered, all plugins are converted into the new scheme. The number of commits is big, but the changes are contained. The last patch is bigger, but also not a big deal, just moving code around.

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -21,9 +21,38 @@ from filelock import FileLock
 from natsort import natsorted
 
 from klpbuild.klplib import utils
+from klpbuild.klplib.cmd import add_arg_lp_name, add_arg_lp_filter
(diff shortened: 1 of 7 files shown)


marcosps:
TBH there were already there and I though that for now we could just migrate the extractor class to a plugin, but you are right. Since everything is changing in this I believe that it's the right time to fix these reports.

fgyanz (approved):
Looks great, thanks!

klpbuild/plugins/extract.py
```diff
@@ -283,7 +283,7 @@ def group_equal_files(lp_name, working_cs):
 
     cs_files = get_cs_code(lp_name, working_cs)
     toprocess = list(cs_files.keys())
-    while len(toprocess):
+    while len(toprocess) > 0:
```
vmezzela, line 286: Just a nit, I think the fix here would have been:
```py
    while toprocess:
```

But no need to change

vmezzela (approved):
Looks good to me!

PR #132, opened 2025-06-30
klplib: Support new MICRO rt-kernels versioning

fgyanz:
klp-build works under the assumption that each kernel version is unique and corresponds to one single codestreams. However, from now on, that won't be the case for newer Micro rt-kernels, that will have the same version as their non-rt counterparts. Luckily, this kernels will still have a unique tag in kernel-source.git and kernel.git: "rpm-$kernel-rt", and "rpm-$kernel" for the non-rt.

Within klp-build the above idea will be used too, and any new Micro rt-kernel version will have the "-rt" suffix appended to it (e.g. "6.4.0-25-rt"). By doing this we keep the uniqueness of the versions.

There is no way to put it kindly, but this workaround is as nasty as it can get. Unfortunately, other solutions required re-working much more codebase and the effort was not worth IMHO.


marcosps (commented):
LGTM, thanks!

PR #133, opened 2025-07-04
setup: Make --cve cmdline option mandatory

fgyanz:
The --cve option is not mandatory in the setup command simply due to testing convenience, breaking the consistency with the rest of commands.

From now on --cve will be mandatory in setup, except when using --no-check option.


marcosps (approved):
LGTM, thanks!

PR #136, opened 2025-07-14
klplib/ksrc: Improve handling of multi-file commits

fgyanz:
klp-build has some trouble processing commits affecting hundreds of patches. Such commits usually get ignored because they are mostly about re-naming or re-organizing the repository tree. But in certain cases, though extremely uncommon, they might be the only way of tracking some patches, and must therefore be kept during the heuristic phase.


marcosps (approved):
LGTM!

PR #137, opened 2025-07-17
tests: Set --no-check option when no CVE is needed

fgyanz:
Some tests were failing because they were still following the old policy


vmezzela (changes requested):
Would you mind changing also `test_templ_ibt_without_externalized_vars`?

fgyanz:
> Would you mind changing also `test_templ_ibt_without_externalized_vars`?

Already did, but I force pushed, so maybe you missed it?

marcosps (approved):
LGTM!

vmezzela:
Oh yes sorry, I had already fetched the PR :D 

So , looks good to me!

PR #139, opened 2025-07-21
Misc

vmezzela:
I've pushed here some misc changes that perhaps are worth considering to be pulled in.

I'm not entirely sure whether `requirements.txt` and `klp-build` are used somewhere I am not aware of, but if that's the case and there are no reasons to have those we can just remove them.


fgyanz (approved):

marcosps:
The klp-build python script was created to be able to run klp-build at the source dir, without having to install it. But I believe that we are thje only three people using it, and we install it using pipx nowadays :)

PR #140, opened 2025-07-21
Drop TemplateGen class

vmezzela:
This is based on #138, when we merge that one I make this PR as ready to merge.

PR #142, opened 2025-07-29
Refactor codestream

vmezzela:
This pull request was to refactor the `Codestream` class, but as a side effect I also cleaned other portions of the code.

There are no logical changes, it's mostly about dropping unneeded fields in favor of methods.

I modified the constructor of the `Codestream` class so that we can avoid dealing with random string like `"12.5rtu4"` but rather use the appropriate class with its functions, like:

```py
cs = Codestream("12.5rtu4")
cs.get_config_content()
```

Also, I dropped `no_check` from `scan` as it was anyway not used and it didn't make sense to have it.

I am opening this PR as Draft as I haven't yet fully reviewed it and still depends on #140 .

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -23,10 +23,11 @@ class Codestream:
         match = re.search(r"(\d+)\.(\d+)(rt)?u(\d+)", name)
         if not match:
             raise ValueError("Name format error")
+        assert match.group(3) in (None, "rt")
 
         self.sle = int(match.group(1))
         self.sp = int(match.group(2))
-        self.rt = match.group(3)
+        self.rt = match.group(3) or ""
         self.update = int(match.group(4))
 
         self.is_micro = self.sle == 6
@@ -175,8 +176,7 @@ class Codestream:
         Return the base codestream name, optionally including 'rt' if it's a
         real-time kernel.
         """
-        rt = "rt" if self.rt else ""
-        return f"{self.sle}.{self.sp}{rt}"
+        return f"{self.sle}.{self.sp}{self.rt}"
 
 
     def full_cs_name(self):
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -11,9 +11,8 @@ from klpbuild.klplib.utils import ARCH, get_workdir, is_mod, get_all_symbols_fro
 from klpbuild.klplib.kernel_tree import init_cs_kernel_tree, file_exists_in_tag, read_file_in_tag
 
 class Codestream:
-    __slots__ = ("sle", "sp", "update", "rt", "ktype", "is_micro",
-                 "project", "patchid", "kernel", "archs", "files", "modules",
-                 "repo", "configs")
+    __slots__ = ("sle", "sp", "update", "rt", "is_micro", "project", "patchid",
+                 "kernel", "archs", "files", "modules", "repo", "configs")
 
     def __init__(self, sle, sp, update, rt, project="", patchid="", kernel="",
                  archs=None, files=None, modules=None, configs=None):
@@ -21,7 +20,6 @@ class Codestream:
         self.sp = sp
         self.update = update
         self.rt = rt
-        self.ktype = "-rt" if rt else "-default"
         self.is_micro = sle == 6
         self.project = project
         self.patchid = patchid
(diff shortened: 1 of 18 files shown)


klpbuild/plugins/setup.py
```diff
@@ -187,7 +187,7 @@ def setup_project_files(lp_name, codestreams, ffuncs, archs):
 
             # Validate if the module being livepatched is supported or not
             if utils.check_module_unsupported(utils.ARCH, mod_path):
-                logging.warning("%s (%s}): Module %s is not supported by SLE", cs.name(), cs.kernel, mod)
```
marcosps, line 190: While at it, would you mind fixing the usages of {} inside the logging messages? pylint complains about it :)

klpbuild/klplib/codestream.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, name, project="", patchid="", kernel="",
         self.project = project
         self.patchid = patchid
         self.kernel = kernel
-        self.archs = archs if archs is not None else []
+
+        self.archs = archs if archs is not None else self.__get_default_archs()
```
marcosps, line 38: Good idea! In the past we set the archs later on in the process, even before the Codestream class, but this is much better!

klpbuild/klplib/utils.py
```diff
@@ -219,26 +219,24 @@ def check_module_unsupported(arch, mod_path):
 
 
 def filter_codestreams(lp_filter, cs_list, verbose=False):
-    full_cs = copy.deepcopy(cs_list)
```
marcosps: There was a catch here in the past, which I don't remember anymore what is was. Maybe there were parallel processes acessing and changing cs_list, but the code changed so much in the last year that I cannot say if the issue still exist.

Let's make it simpler and see if there is an issue still present!

klpbuild/plugins/setup.py
```diff
@@ -10,6 +10,7 @@
 from klpbuild.klplib import utils
 from klpbuild.klplib.cmd import add_arg_lp_name, add_arg_lp_filter
 from klpbuild.klplib.codestreams_data import get_codestreams_data, set_codestreams_data, store_codestreams
+from klpbuild.klplib.supported import get_supported_codestreams
```
marcosps, line 13: The idea of no_check was introduced when we planned to support creating livepatch for any function in the kernel. Since that ship sailed a long time ago, and we only care about CVEs, and we expect them to be backported before creating the LP, I would say that we can also drop it from setup. But we can do that in a different PR.

marcosps:
I liked the changes, thanks a lot for it! I only have small nits, but otherwise LGTM!

klpbuild/plugins/setup.py
vmezzela: Done!

klpbuild/klplib/codestreams_data.py
```diff
@@ -17,12 +17,11 @@
 class CodestreamData:
     cve: str
     archs: list[str]
-    patched_kernels: list[str]
```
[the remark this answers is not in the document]
vmezzela: Yes, though currently it's no longer needed as we anyway save the list of `patched_cs`, that's why I dropped it

vmezzela:
> I liked the changes, thanks a lot for it! I only have small nits, but otherwise LGTM!

Thanks for the quick review!

klpbuild/klplib/utils.py
fgyanz: >  I think now the issue shouldn't persist because the filter_codestreams is always called at where concurrency is not yet a thing

Just a small remark about this. I  have a working implementation where I'm accessing scan() from multiple threads. That is, analyzing several CVE's at the same time. There are several ways to handle the issue, so it's not really a big deal. However, if possible, I'd make every function in `utils.py` re-entrant and thread-safe to avoid future headaches and nasty bugs.

fgyanz:
Looking great! Just one small remark. Thanks :)

klpbuild/klplib/utils.py
vmezzela: `scan` should allocate its own list of codestreams on top of which it calls `filter_codestreams`, so each instance of `scan` should work on its own data.

I dropped that `deepcopy` because the function does not even modify `cs_list` and it allocates two new local lists `result` and `filtered`. Hence from the point of view of this function, its behavior with regard to concurrency shouldn't be affected by this PR, given that the same instance of `cs_list` is never touched by other thread concurrently.

I am assuming you are calling `scan()` from many thread, instead in case you are spawning multiple threads from within `scan()` that would be different and further changes might be required in this regard.

fgyanz: Looks good, thanks! :partying_face:

vmezzela:
I adjusted the logging level of `supported.py`, we don't need to always print such info, but if needed for some reasons we have `--verbose` to enable it.

PR #144, opened 2025-08-06
klplib: utils: Remove support to read symbols from compressed modules

marcosps:
It's very unlikely that klp-ccp will support reading compressed modules like clang-extract. If that would ever happened, we could stop decompressing modules on the setup phase. But for now, all modules are extracted, and even the gzipped configs are extracted, so just read the uncompressed files.

If even klp-ccp allows reading compressed files, this changes can be reverted.


vmezzela (approved):
lgtm!

PR #145, opened 2025-08-06
Pylint improvements

marcosps:
@vmezzela there is a fix related to your refactor on cs-diff plugin. This PR is based on another PR :)

### DIFF
--- a/klpbuild/klplib/ibs.py
+++ b/klpbuild/klplib/ibs.py
@@ -49,7 +49,7 @@ def convert_cs_to_prj(cs, prefix):
     return prefix + "-" + cs.full_cs_name().replace(".", "_")
 
 
-def do_work(func, args):
+def do_work(func, args: list[RPMData]):
     if len(args) == 0:
         return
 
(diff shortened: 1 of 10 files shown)


fgyanz (approved):
LGTM!

klpbuild/klplib/ibs.py
```diff
@@ -19,12 +20,27 @@
...
 from klpbuild.klplib.utils import ARCH, get_all_symbols_from_object, get_datadir
 
 logging.getLogger("osctiny").setLevel(logging.WARNING)
 
 
+# Dataclass for storing rpm related data
+@dataclasses.dataclass
+class RPMData:
+    index: int
+    osc: Osc
```
vmezzela, line 34: Do we really need to keep passing `osc` here? It's not an RPM data in any way. Additionally, it looks like we allocate it just to pass through some functions that eventually use it. Can't we just allocate it where it's used?

marcosps: In the past there were issues regarding too many connection being held at the same time to IBS, so I added the Osc attribute for the IBS class. Maybe there were different reasons why this wasn't working, but this is safe right now at least :)

Maybe we can just drop it and see if it crashes or not? I'm not sure..

vmezzela:
lgtm!

PR #154, opened 2025-10-06
Automation (1/3): Scan: Report modified functions

fgyanz:
Hi guys,

With this PR the `scan` command is now able to report the modified functions of each affected file:

```
6.0rtu2-8 6.0u2-7 15.3u48-58 15.4u31-41 15.5u19-27 15.6rtu2-10 15.6u4-11 15.7rtu0 15.7u0:
FILE: net/sched/sch_ets.c
CONF: CONFIG_NET_SCH_ETS
OBJ: net/sched/sch_ets
FUNCS: ets_qdisc_change, ets_qdisc_dequeue, ets_qdisc_reset, ets_class_qlen_notify
```

The modified functions are found by:
1. Locating the commit introducing the patch in kernel.git. This is done by greping for the `subject` field.
2. Parsing the content/diff of each modified file in the commit, with full function context.

Limitations:
- There is no noticeable slowdown, except in some few cases, due to [this code]( https://github.com/SUSE/klp-build/compare/devel...fgyanz:scan-funcs?expand=1#diff-6e296f3d424cade938c4e89b0eb23e20dc210d23a0d88d82a0c9b6159aa2aa4bR116-R126). Worst case I've seen is around 1 minute. I think it's within an acceptable margin, and we can probably improve it.
- False-positives? False-negatives? None so far, but I'm sure there must be some. For the record, I've tested around 30 CVEs, and with the last iteration 0 issues.

Please, let me know your thoughts. I have a second PR almost ready where the `setup` is fully automated based on the scan report. Meaning that the `--file-funcs`, `--config` and `--module` parameters will be optional and only meant for corner cases were the scan report is wrong.

### DIFF
--- a/klpbuild/klplib/file2config.py
+++ b/klpbuild/klplib/file2config.py
@@ -33,11 +33,14 @@ def _find_config(cs, base_dir, relative_obj_path, deep):
     if deep > 10:
         return None, ""
 
+    if Path(".") == base_dir:
+        return "CONFIG_SUSE_KERNEL", "vmlinux"
+
     make_file = Path(base_dir, "Makefile")
 
     lines = _load_makefile(cs, make_file)
 
-    if not lines and Path(".") != base_dir:
+    if not lines:
         relative_obj_path = base_dir.name + "/" + relative_obj_path
         return _find_config(cs, base_dir.parent, relative_obj_path, deep+1)
 
@@ -48,6 +51,9 @@ def _find_config(cs, base_dir, relative_obj_path, deep):
 
         # target found, check if this one with config
         target = sep[0]
+        if target.startswith('obj-y'):
+            # If it's built-in then check the config of the parent directory
+            return _find_config(cs, base_dir.parent, base_dir.name + '/', deep)
         if target.startswith('obj-'):
             return _sanitize_config(target), str((base_dir/relative_obj_path).with_suffix(''))
 
@@ -70,18 +76,19 @@ def find_configs_for_files(cs, file_paths: list):
     missing = []
 
     if not file_paths:
-        return configs, build_ins, missing
+        return configs, missing
 
     for path in file_paths:
         path = path.strip()
+        # Do not check headers
+        if path.endswith('h'):
+            continue
         obj_file = Path(path.replace('.c', '.o'))
         config, obj = _find_config(cs, obj_file.parent, obj_file.name, 0)
         if not config:
             missing.append(path)
-        elif config == 'y' or config == 'm':
-            build_ins.append(path)
         elif config.startswith('CONFIG_'):
             configs[path] = {'config': config, 'obj': obj}
         # else there is garbage like 'subst', 'vds' for wrongly parsed input
 
-    return configs, build_ins, missing
+    return configs, missing
--- a/klpbuild/klplib/kernel_tree.py
+++ b/klpbuild/klplib/kernel_tree.py
@@ -7,6 +7,7 @@ import logging
 import os
 import shutil
 import subprocess
+import re
 
 from multiprocessing import Lock
 from functools import wraps
@@ -56,7 +57,8 @@ def __fetch_kernel_tree_tags():
         logging.info("Failed to update kernel tree tags\n%s", ret.stderr)
 
 
-# Currently this function returns the date of the patch and its subject
+# Currently this function returns the date of the patch, its subject
+# and the hash of the corresponding commit in kernel-source.
 @__check_kernel_tags_are_fetched
 def get_commit_data(commit, savedir=None):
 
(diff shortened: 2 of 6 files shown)


vmezzela (changes requested):
Hi Fernando, thanks for the PR! I left come proposed changes inline. Feel free to adopt/discard them

klpbuild/klplib/file2config.py
```diff
@@ -74,14 +80,15 @@ def find_configs_for_files(cs, file_paths: list):
```
vmezzela, line 80: At line 79 here (couldn't see it from the diff context), we're returning 3 variables instead of 2.

klpbuild/klplib/file2config.py
```diff
@@ -74,14 +80,15 @@ def find_configs_for_files(cs, file_paths: list):
...
         if not config:
             missing.append(path)
-        elif config == 'y' or config == 'm':
-            build_ins.append(path)
         elif config.startswith('CONFIG_'):
             configs[path] = {'config': config, 'obj': obj}
         # else there is garbage like 'subst', 'vds' for wrongly parsed input
 
-    return configs, build_ins, missing
+    return configs, missing
```
vmezzela, line 94: Also `missing` is no longer used, but perhaps is still worth to keep it if we ever need it in the future?

klpbuild/klplib/patch.py
```diff
@@ -32,20 +32,21 @@ def analyse_files(cs_list, sle_patches):
             bc = cs.base_cs_name()
             patches = sle_patches[bc]
             branch = KERNEL_BRANCHES[bc]
-            files = get_patch_files(patches, branch)
-            fconfigs, _, missing = find_configs_for_files(cs, files)
+            files_funcs = get_patches_files(patches, branch)
+            files, _ = find_configs_for_files(cs, files_funcs.keys())
```
vmezzela, line 36:
```suggestion
            files_config, _ = find_configs_for_files(cs, files_funcs.keys())
```
What about renaming this? Even with another name that makes it clear that the dict contains the config and object.

klpbuild/klplib/patch.py
```diff
@@ -32,20 +32,21 @@ def analyse_files(cs_list, sle_patches):
...
 
-            for file in missing:
-                key = f"{file}::"
-                if cs not in report[key]:
-                    report[key].append(cs)
+            for file, funcs in files_funcs.items():
+                fdat = files.get(file, {})
```
vmezzela, line 39: What about slightly restructuring the code here?

This code is precomputing some values and then storing them in dictionaries of sets before the loop begin. Then in the loop such dicts (see `files_funcs`, `files`) are accessed using the `file` as key to retrieve the item computed before. Time-wise, this is not an issue, but it's just not easy to follow, especially because when reading is not clear what each object is (for instance, we might forget about the empty dict returned by `get_patches_files()` when we will touch this code in the future).

Wouldn't it be just enough to get the list of files modified by the patches, and then while iterating over them we retrieve the modified funcs and affected configs/objs?



```py
    for cs in cs_list:
            bc = cs.base_cs_name()
            patches = sle_patches[bc]
            branch = KERNEL_BRANCHES[bc]

            for file in get_patches_files(patches, branch) :
                funcs = get_funcs_for_file(patches, branch, file)
                configs = find_configs_for_file(cs, file) # fdat?
                ...
```

Please note that this requires modifying `find_configs_for_fileS()` to `find_configs_for_file()`, and splitting `get_patches_files()` into `get_funcs_for_files()`, so that the former only returns the list of files and the latter the list of funcs for each file modified by the patches.

But tbh I don't have the big picture here so I'm not sure if with this approach we lose something

vmezzela (changes requested):

klpbuild/klplib/kernel_tree.py
```diff
@@ -78,6 +78,58 @@ def get_commit_data(commit, savedir=None):
...
+
+    ret = subprocess.run(["git", "-C", kernel_tree, "show", "-W",
+                          "--pretty='%b'", f"{commit}", f"{file_path}"],
+                         check=False, capture_output=True, text=True)
+    return ret.stdout
+
+
+@__check_kernel_tags_are_fetched
+def find_commit(subject, branch):
+    """
```
marcosps: This was already tried in the past, and had a couple of corner cases. I don't recall anymore, since it was a long time ago, but let's see how this one behaves.

marcosps: Ah! You're using `kernel-sources` and git-show with `-W`, so yeah, this might work!!!

marcosps (approved):
The changes looks sane. Thanks for working on it!

klpbuild/klplib/kernel_tree.py
fgyanz: Yep, though the git-show is in `kernel.git`. 
it's been working really nicely so far! I have only found one corner case that I'm unable to cover with the current solution: CVE 2024-53168.

```
15.3u49-57:
FILE: net/core/sock.c
CONF: CONFIG_NET
OBJ: net/core
FUNCS: release_proto_idx, sock_inuse_get

15.4u32-40 15.5u20-26:
FILE: net/core/sock.c
CONF: CONFIG_SUSE_KERNEL
OBJ: vmlinux
FUNCS: release_proto_idx, sock_inuse_get
```
Those functions are false-positives, they are actually not touched by the patch. This is the commit diff:
```
@@ -3284,11 +3282,6 @@ int sock_prot_inuse_get(struct net *net, struct proto *prot)
 }
 EXPORT_SYMBOL_GPL(sock_prot_inuse_get);

-static void sock_inuse_add(struct net *net, int val)
-{
-       this_cpu_add(*net->core.sock_inuse, val);
-}
-
 int sock_inuse_get(struct net *net)
 {
        int cpu, res = 0;
@@ -3367,9 +3360,6 @@ static inline void release_proto_idx(struct proto *prot)
 {
 }

-static void sock_inuse_add(struct net *net, int val)
-{
-}
 #endif
```
This one is pretty tricky, both `sock_inuse_get` and `release_proto_idx` fool our regexs. I'm not sure how to workaround it without messing other cases.

klpbuild/klplib/file2config.py
fgyanz: done

klpbuild/klplib/file2config.py
fgyanz: Agreed, I had no plans of getting rid of it :)

klpbuild/klplib/patch.py
fgyanz: done

klpbuild/klplib/patch.py
fgyanz: I agree, the code is quite ugly, but I'm afraid it's not that easy. Initially I did try to implement your approach, but did not work. As I see it, pre-computing the values is mandatory if you don't want to fall into a loop nesting trap. For each patch you need to check its files, and for each file, its functions. And then you need to unify all the patches data to create a unique key `config-module-file-functions` per codestream.
I tried improving the naming, let me know what you think!

fgyanz (commented):
Hi guys! Hopefully last update! I fixed some bugs and attended your comments.

marcosps:
I did another look, and things look good. Thanks!

vmezzela:
lgtm!

PR #155, opened 2025-10-06
Simple cache implementation

vmezzela:
A few days ago I had issues while using klp-build due to `git fetch` being really slow. Hence I implemented this simple cache to avoid fetching every time.

Please let me know if that's something that you think is worth merging. If so, I will reiterate over the proposed PR to see if there's something that can be done better.

One thing that might be added is a global `--no-cache` option that disables the cache.


vmezzela:
> Well, IMHO, this is not really a problem that happens everyday, so I'm not sure if this would be used in the end, but I understand that it can make someone's day product, even if this happens once in a year :)

Are you referring to `git fetch` being slow?

Because the PR no longer address that *issue* only, but also others. As for my previous comment it also is useful when running a command that does an expensive computation and we need to rerun it again to adjust some parameters (I mentioned running setup/extract on a livepatch that affects vmlinux): with this feature, that expensive computation will just be cached making it all faster.

But as said, this is just a POC, we should decide where to use the decorator and where not.

PR #156, opened 2025-10-09
Automation (2/3): setup

fgyanz:
From now on, setup will no longer require more parameters other than '--cve' and '--name'. The livepatch technical information (that is: files, functions, configs, modules and archs) will be directly retrieved from the report generated by scan(). The user will only need to double-check that the livepatch information is actually correct. In case it is not,
there's still as a fallback option the manual input ('--conf', '--file-func', etc).

This PR depends on https://github.com/SUSE/klp-build/pull/154.

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -348,39 +348,6 @@ class Codestream:
 
         return configs
 
-    def validate_config(self, archs, conf, mod):
-        configs = {}
-        cs_config = self.get_all_configs(conf)
-
-        # Validate only the specified architectures, but check if the codestream
-        # is supported on that arch (like RT that is currently supported only on
-        # x86_64)
-        for arch in archs:
-            # Check if the desired CONFIG entry is set on the codestreams's supported
-            # architectures, by iterating on the specified architectures from the setup command.
-            if arch not in self.archs:
-                continue
-
-            try:
-                conf_entry = cs_config.pop(arch)
-            except KeyError as exc:
-                raise RuntimeError(f"{self.full_cs_name()}: {conf} not set on {arch}. Aborting") from exc
-
-            if conf_entry == "m" and mod == "vmlinux":
-                raise RuntimeError(f"{self.full_cs_name()}:{arch} ({self.kernel}): Config {conf} is set as module, but no module was specified")
-            if conf_entry == "y" and mod != "vmlinux":
-                raise RuntimeError(f"{self.full_cs_name()}:{arch} ({self.kernel}): Config {conf} is set as builtin, but a module {mod} was specified")
-
-            configs.setdefault(conf_entry, [])
-            configs[conf_entry].append(f"{self.full_cs_name()}:{arch}")
-
-        # Validate if we have different settings for the same config on
-        # different architecures, like having it as builtin on one and as a
-        # module on a different arch.
-        if len(configs.keys()) > 1:
-            print(configs["y"])
-            print(configs["m"])
-            raise RuntimeError(f"{self.full_cs_name()}: Configuration mismatach between codestreams. Aborting.")
 
     def find_obj_path(self, arch, mod):
         # Return the path if the modules was previously found for ARCH, or refetch if
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -312,6 +312,13 @@ class Codestream:
     def get_mod(self, mod):
         return self.modules[mod]
 
+
+    def get_file_mod(self, file, arch=ARCH):
+        fdat = self.files[file]
+        conf_arch = self.configs[fdat["conf"]][arch]
+        return "vmlinux" if conf_arch == 'y' else fdat["module"]
+
+
     def is_module_supported(self, mod):
         return ksrc_is_module_supported(mod, self.kernel)
 
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -6,6 +6,7 @@
 import re
 import subprocess
 import tempfile
+import logging
 
 from pathlib import Path, PurePath
 from importlib import resources
@@ -169,16 +170,19 @@ class Codestream:
         # We support all architecture for all other codestreams
         return ["x86_64", "s390x", "ppc64le"]
 
-
     def set_files(self, files):
         self.files = files
 
+    def set_configs(self, configs):
+        self.configs = {conf: self.get_all_configs(conf) for conf in configs}
+
+    def set_archs(self, archs):
+        self.archs = list(set(archs) & set(self.__get_default_archs()))
 
     def get_kernel_type(self, suffix=False):
         dash = "-" if suffix else ""
         suffix = "rt" if self.rt else "default"
         return dash + suffix
-
 
     def get_full_kernel_name(self):
         """Returns the kernel name with flavor suffix"""
@@ -323,6 +327,10 @@ class Codestream:
         the configuration is not set the return value will be an empty dict.
         """
         configs = {}
+
+        if conf and not conf.startswith("CONFIG_"):
+            logging.error(f"Invalid config '{conf}': Missing CONFIG_ prefix")
+            sys.exit(1)
 
         for arch in self.archs:
             kconf = self.get_config_content(arch)
--- a/klpbuild/klplib/patch.py
+++ b/klpbuild/klplib/patch.py
@@ -40,9 +40,9 @@ def analyse_files(cs_list, sle_patches):
             for file, funcs in files_funcs.items():
                 conf = files_conf.get(file, {})
                 if conf:
-                    cs.files[file] = {'symbols': funcs}
+                    cs.files[file] = {'symbols': list(funcs)}
                     cs.files[file].update(conf)
-                    key = f"{file}:{conf['config']}:{conf['obj']}:{sorted(funcs)}"
+                    key = f"{file}:{conf['conf']}:{conf['module']}:{sorted(funcs)}"
                 else:
                     key = f"{file}:::"
 
@@ -64,8 +64,8 @@ def print_files(report):
             logging.info(f"{cs_str}:\nFILE: {file}\n")
             continue
 
-        conf = cs.files[file]['config']
-        obj = cs.files[file]['obj']
+        conf = cs.files[file]['conf']
+        obj = cs.files[file]['module']
         funcs = cs.files[file]['symbols']
         logging.info("%s:\nFILE: %s\nCONF: %s\nOBJ: %s\nFUNCS: %s\n",
                      cs_str, file, conf, obj, ', '.join(funcs))
@@ -86,7 +86,7 @@ def analyse_configs(cs_list):
     report = defaultdict(list)
 
     for cs in cs_list:
-        configs = {dat['confing'] for _, dat in cs.files.items()}
+        configs = {dat['conf'] for _, dat in cs.files.items()}
         cs.set_configs(configs)
         for conf, archs in cs.configs.items():
             key = f"{conf}:{archs}"
@@ -150,8 +150,8 @@ def analyse_kmodules(cs_list):
 
     for cs in cs_list:
         for _, dat in cs.files.items():
-            conf = dat['config']
-            mod = dat['obj']
+            conf = dat['conf']
+            mod = dat['module']
             if mod in cs.modules:
                 continue
 
@@ -191,6 +191,9 @@ def filter_unsupported_kmodules(cs_list):
         supported = [s for _, s in cs.modules.items() if s]
         if not supported:
             unset_cs.append(cs)
+
+        # Cleanup for future re-use
+        cs.modules.clear()
 
     for cs in unset_cs:
         cs_list.remove(cs)
--- a/klpbuild/klplib/patch.py
+++ b/klpbuild/klplib/patch.py
@@ -86,14 +86,10 @@ def analyse_configs(cs_list):
     report = defaultdict(list)
 
     for cs in cs_list:
-        for _, dat in cs.files.items():
-            conf = dat['config']
-            if conf in cs.configs:
-                continue
-
-            cs.configs[conf] = cs.get_all_configs(conf)
-
-            key = f"{conf}:{cs.configs[conf]}"
+        configs = {dat['confing'] for _, dat in cs.files.items()}
+        cs.set_configs(configs)
+        for conf, archs in cs.configs.items():
+            key = f"{conf}:{archs}"
             if cs not in report[key]:
                 report[key].append(cs)
 
--- a/klpbuild/klplib/utils.py
+++ b/klpbuild/klplib/utils.py
@@ -225,6 +225,16 @@ def filter_codestreams(lp_filter, cs_list, verbose=False):
 
     return result
 
+
+def affected_archs(cs_list):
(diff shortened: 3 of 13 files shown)


vmezzela (changes requested):
Hi @fgyanz , thanks for the PR! This is really useful! I left some comments inline.

My only concern is related to the `cs.configs` and `cs.files` field. These are becoming more than just normal fields in the Codestream object. Furthermore, they contain overlapping data.

- `cs.configs` : `{'CONFIG_SUSE_KERNEL': {'x86_64': 'y', 's390x': 'y'}}`
- `cs.files` : `{'fs/namespace.c': {'conf': 'CONFIG_SUSE_KERNEL', 'module': 'vmlinux', 'symbols': ['clone_private_mount']}}`

I think we can merge these two fields. 
What do you think about creating a new class for this purpose?

In this way we avoid having objects that are nested dictionaries, and it would be more clear (by reading the code) what the field contains.

klpbuild/plugins/setup.py
```diff
@@ -145,64 +150,65 @@ def setup_codestreams(lp_name, data):
...
-        cs.set_files(copy.deepcopy(ffuncs))
-
         # Check if the files exist in the respective codestream directories
-        mod_syms = {}
         for f, fdata in cs.files.items():
-
             mod = fdata["module"]
-            cs.validate_config(archs, fdata["conf"], mod)
+            conf = fdata["conf"]
+            archs = cs.configs[conf]
```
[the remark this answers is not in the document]
fgyanz: Erf sorry if it's not clear. Those are the values returned from file2config library. `y` stands for builtin (so vmlinux.o)
and `m` for module (so whatever module is associated with that CONFIG). Both enable that CONFIG for one specific arch. I've written a helper function for addressing this obscurity, [get_file_mod](https://github.com/SUSE/klp-build/pull/156/files#diff-c1152f999d940b753edcc0565518c984956739d1bb96e1b7bb37d61fb2f8d665R283). 

TBH I don't think it's that obscure. But sure, there are probably better ways and data structures that can be used here. Beware that reworking this is no minor task, both variables are used in many places. I'd vote for discussing possible ideas privately and once we come up with a good solution, reworking the affected code in a different PR.

fgyanz:
Thanks for comments @vmezzela, I agree with most of them. Just a remark:

> My only concern is related to the cs.configs and cs.files field. These are becoming more than just normal fields in the Codestream object. Furthermore, they contain overlapping data.
> 
>     cs.configs : {'CONFIG_SUSE_KERNEL': {'x86_64': 'y', 's390x': 'y'}}
>     cs.files : {'fs/namespace.c': {'conf': 'CONFIG_SUSE_KERNEL', 'module': 'vmlinux', 'symbols': ['clone_private_mount']}}
> 
> I think we can merge these two fields.
> What do you think about creating a new class for this purpose?
> 
> In this way we avoid having objects that are nested dictionaries, and it would be more clear (by reading the code) what the field contains.

I answered it [here](https://github.com/SUSE/klp-build/pull/156#discussion_r2473080454)

marcosps:
If the file/funcs cannot be found by the setup phase/scan, I think that we should error out, otherwise later on we can call extract command with an empty file-funcs on `codestreams.json`. What do you think @fgyanz ?

marcosps:
> If the file/funcs cannot be found by the setup phase/scan, I think that we should error out, otherwise later on we can call extract command with an empty file-funcs on `codestreams.json`. What do you think @fgyanz ?

 At least this happened with a patch file that only patches a header, and so the file2config fails, and there is not way for us to know which C file we need to use in order to extract the function.

klpbuild/plugins/setup.py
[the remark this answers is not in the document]
vmezzela: > Erf sorry if it's not clear. Those are the values returned from file2config library. `y` stands for builtin (so vmlinux.o) and `m` for module (so whatever module is associated with that CONFIG).

Sure, it was clear after printing the object. But if one comes here after a while not reading the code I think it's not easy to understand it.

What I meant is just that I'd rather pack it into a dedicated data structure so it's easier to debug in the future because we'd know what contains what by looking ad the definition of the class. By reading the code it's just not easy to understand as it is no longer a simple object but nested dictionaries/lists.

> I'd vote for discussing possible ideas privately and once we come up with a good solution, reworking the affected code in a different PR.

Sure, I just think that the later we address this and the more complex would be to adjust the code. But on the other hand I don't want it to be a stopper and most likely adjusting and then rebasing this PR is even more a pain so nevermind.. we'll try to fix it afterward.

fgyanz:
Yup, agreed. Actually, good news, I already have implemented that, but haven't pushed it just yet. It was a needed during CI development to catch corner cases, as you the one you mentioned. Hopefully you'll find it useful :)

klpbuild/plugins/setup.py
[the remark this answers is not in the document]
fgyanz: Right, I understand your concerns. I've been more focused improving functionality rather than code quality in the last PRs. Once this batch of PRs gets merged and released, it will be the perfect opportunity to cleanup the code and fix bugs IMO. I created a issue to keep track of this: https://github.com/SUSE/klp-build/issues/164

klpbuild/klplib/codestream.py
```diff
@@ -290,6 +293,9 @@ def get_all_configs(self, conf):
         """
         configs = {}
 
+        if conf and not conf.startswith("CONFIG_"):
+            raise ValueError(f"Invalid config '{conf}': Missing CONFIG_ prefix")
```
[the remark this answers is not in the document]
fgyanz: done

klpbuild/plugins/scan.py
```diff
@@ -115,6 +115,10 @@ def scan(cve, conf, lp_filter, download, savedir=None):
     unaffected_cs = []
     for cs in filtered_codesteams:
 
+        cs.set_archs(archs)
+        if not cs.archs:
+            continue
```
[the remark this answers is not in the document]
fgyanz: yeah, if you filter `--arch s390x`, for example, the `rt` kernels will have an empty `cs.archs`.

klpbuild/plugins/setup.py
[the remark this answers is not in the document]
vmezzela: Thanks a lot!

klpbuild/plugins/scan.py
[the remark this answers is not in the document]
vmezzela: Oh I see, thanks for clarifying! :)

klpbuild/plugins/scan.py
```diff
@@ -115,6 +115,10 @@ def scan(cve, conf, lp_filter, download, savedir=None):
     unaffected_cs = []
     for cs in filtered_codesteams:
 
+        cs.set_archs(archs)
+        if not cs.archs:
```
marcosps, line 120: Maybe add a comment saying the set_archs will only set the archs that the codestream has support to, and for this arch=ppc64 in a MICRO codestream will return empty. Not really a big deal, but I had to go back and check what set_archs does to understand this check.

klpbuild/plugins/setup.py
```diff
@@ -149,60 +160,71 @@ def setup_codestreams(lp_name, data):
...
     # Setup the missing codestream info needed
     for cs in codestreams:
-        cs.set_files(copy.deepcopy(ffuncs))
-
-        # Check if the files exist in the respective codestream directories
-        mod_syms = {}
+        # Check if the files and symbols exist in the respective codestream directories
         for f, fdata in cs.files.items():
+            conf = fdata["conf"]
+            assert conf
```
marcosps: Maybe this assert could be added in set_confs? I don't believe that we have a reason to continue with the procedure if a conf couldn't be found.

marcosps:
I liked the changes. There is a small nit, but isn't really an issue. For me we are ready to merge this one as well. Thanks!

klpbuild/klplib/codestream.py
```diff
@@ -312,6 +323,10 @@ def get_all_configs(self, conf):
         """
         configs = {}
 
+        if conf and not conf.startswith("CONFIG_"):
+            logging.error(f"Invalid config '{conf}': Missing CONFIG_ prefix")
```
marcosps, line 339: Missing `import logging`

marcosps: Well, shouldn't it be a Raise? Right now it fails a test, since we expect an `ErrorValue`

vmezzela: Hi Marcos, please see the [previous comment](https://github.com/SUSE/klp-build/pull/156#discussion_r2419153437). I suggested not to raise anything but rather just print the error and gracefully exit. I see no reasons to have a stack trace printed in this case...

klpbuild/klplib/patch.py
```diff
@@ -86,14 +86,10 @@ def analyse_configs(cs_list):
...
     for cs in cs_list:
-        for _, dat in cs.files.items():
-            conf = dat['config']
-            if conf in cs.configs:
-                continue
-
-            cs.configs[conf] = cs.get_all_configs(conf)
-
-            key = f"{conf}:{cs.configs[conf]}"
+        configs = {dat['confing'] for _, dat in cs.files.items()}
```
vmezzela: `confing` typo

vmezzela: Okay, fixed in "klplib/patch: Make output compatible with setup "

klpbuild/klplib/codestream.py
marcosps: Ok, I'll adjust the error message in Symbol Validation. But still, the `import logging` is missing.

vmezzela: Agree!

fgyanz: done

klpbuild/plugins/scan.py
fgyanz: done

klpbuild/plugins/setup.py
fgyanz: done

klpbuild/plugins/setup.py
```diff
@@ -145,64 +150,65 @@ def setup_codestreams(lp_name, data):
...
+        logging.warning("%s (%s): Module %s is not supported by SLE",
+                        cs.full_cs_name(), cs.kernel, mod)
+
+    cs.modules[mod] = str(mod_path)
+
+
+def __setup_check_syms(cs, mod, syms, arch):
+    # Verify if the functions exist in the specified object
+    arch_syms = cs.check_symbol_archs(arch, mod, syms, False)
+    if arch_syms:
```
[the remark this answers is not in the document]
fgyanz: Apparently it's not. We need the guard to avoid crashing with `arch_syms.items()`. I slightly changed the code to make it more clear.

klpbuild/plugins/setup.py
```diff
@@ -145,64 +150,65 @@ def setup_codestreams(lp_name, data):
...
+    if utils.check_module_unsupported(utils.ARCH, mod_path):
+        logging.warning("%s (%s): Module %s is not supported by SLE",
+                        cs.full_cs_name(), cs.kernel, mod)
+
+    cs.modules[mod] = str(mod_path)
+
+
+def __setup_check_syms(cs, mod, syms, arch):
+    # Verify if the functions exist in the specified object
+    arch_syms = cs.check_symbol_archs(arch, mod, syms, False)
```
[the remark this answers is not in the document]
fgyanz: > At the point when this function is called, arch should be already set in cs.archs (I mean in the cs instance). So I think that check_symbol_archs should be adjusted accordingly in this patchset.

Right, but note that the `mod` might be different for each arch. AFAICS we need to call `cs.check_symbol_archs` for each arch in order to handle the above case.

fgyanz:
Hi guys! Attended comments, ready for review. Also fixed this issue. https://github.com/SUSE/klp-build/pull/160#issuecomment-3482268337 Turns out the arch filtering was not working when `no-check` was enabled. Thanks @marcosps for reporting it.

marcosps (approved):
LGTM! Thanks a lot!

klpbuild/plugins/setup.py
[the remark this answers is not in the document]
vmezzela: still.. `check_symbol_archs()`  never returns `None`. In case all the symbols are present, it returns an empty dict `{}` and `for arch, syms in arch_syms.items()` would just have no effect and return. Which is exactly the same as using  `if not arch_syms: return`, therefore my proposal to drop it. However, I don't want to bother updating the PR just for this nit.

klpbuild/klplib/utils.py
```diff
@@ -225,6 +225,16 @@ def filter_codestreams(lp_filter, cs_list, verbose=False):
...
     return result
 
+
+def affected_archs(cs_list):
+    conf_archs = {}
+    for cs in cs_list:
+        for val in cs.configs.values():
+            conf_archs.update(val)
+
+    return list(sorted(conf_archs))
```
vmezzela: In this function, after the loop, `conf_arch` is a dictionary like: `{'x86_64': 'y', 's390x': 'y', 'ppc64le': 'y'}`.

However, after calling `sorted()` on it, it returns a list: `['ppc64le', 's390x', 'x86_64']`. Hence no need to call `list()`.

Also, wouldn't it be better to just use a `set()`? I see no point in using a dictionary here. Every time `conf_arch.update(val)` is called it would overwrite the entry with that key.

But again, since this works I wouldn't bother to update it. But, I think this is another reason why we should think about a better data structure or Class to hold such info.

klpbuild/klplib/patch.py
```diff
@@ -192,6 +192,9 @@ def filter_unsupported_kmodules(cs_list):
         if not supported:
             unset_cs.append(cs)
 
+        # Cleanup for future re-use
+        cs.modules.clear()
```
vmezzela, line 196: Just out of curiosity, what is that for? I don't know if it's something that's worth making more clear in the comment above?

klpbuild/plugins/setup.py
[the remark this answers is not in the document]
vmezzela: Right, make sense!

vmezzela (approved):
Hi Fernando,
I left some other comments inline. They are mostly nits, hence if you don't feel like addressing them just feel free to merge it.

Thanks a lot for working on it! it's awesome!! \o/

klpbuild/klplib/utils.py
fgyanz: Ouch good catch! You are right. I wrote this part in a hurry and didn't realize I was using a `dict`, but my intention was to have a `set()`.

klpbuild/klplib/patch.py
fgyanz: An ugly hack, I'm afraid. The content and purpose of this data structure during `analysis` slightly differs from the next phases. As such, to avoid conflicts and discrepancies I decided to empty it and start from scratch once the `analysis` was completed. This is indeed due to poor design, and I strongly agree with a rework.

klpbuild/plugins/setup.py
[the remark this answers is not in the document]
fgyanz: my bad, didn't notice the empty dict. Done

fgyanz:
Thanks for the comments, @vmezzela ! I addressed them, and should be ready again :)

PR #157, opened 2025-10-13
plugins/extract: Apply the patches with git

fgyanz:
The current implementation uses quilt to apply the patches, and although it works in most cases, it lacks the flexibility offered by git. In the past it made sense to use quilt due to several limitations in the tooling, but that's not the case anymore.

From now on, during 'extract' each codestream will have its own working branch, making it easier to track changes in case of errors. From a high level the process will remain the same as before: 

0. run extract()
1. remove patches
2. apply patches
3. run klp-ccp
4. remove patches

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -345,29 +345,26 @@ def group_equal_files(lp_name, working_cs):
         logging.info("\t%s", group)
 
 
-def quilt_log_path(lp_name, apply_patches):
-    if apply_patches:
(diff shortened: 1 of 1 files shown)


fgyanz:
> But I guess that any 15.5u1 patches directory should be created manually or klp-build is able to create it when needed?

klp-build used to create always those directories, but we got rid of that functionality. I'd bring it back but just on a per need basis. Meaning that if `15.5u0` is missing all the patches, then klp-build would just use `15.5` dir.

I think it should be fairly easy to implement, we already have all the information available.

PR #158, opened 2025-10-15
ksrc: Add initial SL16 repositories

marcosps:


vmezzela:
Hi, before we merge this, I wonder how does it work for the realtime livepatches. So there will be no -rt branch on kernel-source.git (and kernel.git), but we will have SLE16-RT_Update_* in kgraft-patches.

Should we have also an entry `16.0rt : SL-16.0` or some other logic so that the realtime livepatch will also be considered?

marcosps:
> Hi, before we merge this, I wonder how does it work for the realtime livepatches. So there will be no -rt branch on kernel-source.git (and kernel.git), but we will have SLE16-RT_Update_* in kgraft-patches.
> 
> Should we have also an entry `16.0rt : SL-16.0` or some other logic so that the realtime livepatch will also be considered?

Makes sense. I'll check why SLE16-RT is not being added to supported.csv. Once I figure that out I'll update this PR.

marcosps:
@fgyanz fixed!

Issue #11, opened 2024-07-30
Multi-file LP support


lhruska:
In case of LPing multiple files, the ccp creates multiple C files with individual `_init` and `_cleanup` functions that must be then manually added to the `livepatch_xxx.h` and called individually in the main `_init` and `_cleanup` function.
Automating this would save a lot of time.

marcosps:
This topic is on my TODO list for quite a while :)

I hope that we could do that soon, so it's even quicker to create a LP!

fgyanz:
To whoever decides to work on this issue: The end result should be quite similar to the one found in [kernel-livepatch](https://github.com/SUSE/kernel-livepatch/blob/master/scripts/register-patches.sh).

fgyanz:
Closing Issue because the multi-file LP support has been added in #160.

PR #160, opened 2025-10-20
Automation (3/3): livepatch extraction

fgyanz:
This PR depend on: #154, #156 and [klp-ccp macro fix](https://github.com/fgyanz/klp-ccp/tree/headers-fix). The latter being a branch that:
- Removes the prologue of low-level macros added to every generated livepatch file. The issue was with the header inclusion heuristics, so now klp-ccp will just include all the headers instead.
- Export a `patched_funcs` file with all the livepatched functions (excluding inlined ones).

Based on the above information, this PR adds the following improvements:
- Declaring the klpp functions in the livepatch header.
- Declaring needed structs in the header.
- Removing the 'static' keyword from the klpp functions definition.
- Exporting the right symbols in "patched_funcs.csv".
- Better support for multi-file livepatches. Wire-up the inits() and cleanups() in the main livepatch.c file. (#11 )
- Adds a worktree for the kgraft-patches repository and a `kgraft` library for interacting with it.
- In the `push` command, instead of cloning the kgraft-patches repository for each
  codestream, use instead the `kgraft` library and have a branch for each codestream.
- Add a new 'commit' command for commiting the extracted livepatches to the kgraft-patches repository.

Take for instance CVE  `2024-53164`. Up until now, the livepatch would have been generated by running:
```
1)   klp-build setup --cve 2024-53164 -n bsc1246019
                   --conf CONFIG_NET_SCH_CAKE --mod-file-funcs sch_cake net/sched/sch_cake.c cake_drop
                   --conf CONFIG_NET_SCH_CHOKE --mod-file-funcs sch_choke net/sched/sch_choke.c choke_drop_by_idx

2) klp-build extract -n bsc1246019 --apply-patches
3) manually create the branches in kgraft-patches and edit the livepatches.
4) klp-build push -n bsc1246019
```

The proposed changes would just require:
```
1) klp-build setup --cve 2024-53164 -n bsc1246019
2) klp-build extract -n  bsc1246019
3) klp-build commit -n bsc1246019 (optionally)
4) klp-build push -n bsc1246019
```


klpbuild/klplib/templ.py
```diff
@@ -482,24 +545,29 @@ def __generate_lp_file(lp_name, lp_path, cs, src_file, out_name):
...
-        temp_str = TEMPL_HOLLOW
+        if cs.needs_ibt():
+            return
+        inits, cleanups = get_multi_funcs(cs, lp_name)
+        tvars.update({"inits": inits, "cleanups": cleanups})
+        temp_str = TEMPL_MULTI_ENTRY
         lp_inc_dir = Path("non-existent")
     else:
         fdata = cs.files[str(src_file)]
+        mod = cs.get_file_mod(src_file)
```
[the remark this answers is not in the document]
vmezzela: Same in `pathced_funcs.csv`

fgyanz:
Thanks @vmezzela ! you are right. I just stumbled upon this same issue in a testing environment. I'll fix it asap :)

klpbuild/klplib/templ.py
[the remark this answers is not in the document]
fgyanz: yep, is that a big deal?  AFAIK it should work all the same, no?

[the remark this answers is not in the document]
vmezzela: At the time I added  this comment I didn't know it would have worked anyway :D 
But actually you are right, so no issue.

klpbuild/plugins/extract.py
```diff
@@ -380,32 +380,74 @@ def group_equal_files(lp_name, working_cs):
...
+    if apply_patches:
+        bname = get_lp_branch(lp_name, cs)
+        err = subprocess.run(["git", "branch", "-D", f"{bname}"], cwd=sdir,
+                             stdout=subprocess.DEVNULL,
+                             stderr=subprocess.PIPE, check=False)
+        if err.returncode != 0 and f"'{bname}' not found" not in str(err.stderr):
+            raise RuntimeError(f"Failed to delete branch {bname}:\n{err.stderr}\n")
 
 
+<<<<<<< Updated upstream
```
vmezzela: leftover?

klpbuild/plugins/extract.py
```diff
@@ -380,32 +380,74 @@ def group_equal_files(lp_name, working_cs):
...
 def apply_all_patches(lp_name, cs, apply_patches):
+=======
+
+def apply_patch(patch, sdir):
```
vmezzela, line 425: can't we just use `patch` here? why do we need to commit?

klpbuild/plugins/extract.py
```diff
@@ -380,32 +380,74 @@ def group_equal_files(lp_name, working_cs):
...
+    subprocess.run(["git", "am", "--continue"],
+                   stdout=subprocess.DEVNULL,
+                   stderr=subprocess.DEVNULL,
+                   cwd=sdir, check=False)
+
+    return True
+
+
+def apply_all_patches(lp_name, cs):
+>>>>>>> Stashed changes
```
vmezzela: leftover

klpbuild/plugins/extract.py
fgyanz: Ah, sorry, I failed to clarify that point in the comments. There are actually several reasons:
- `git-am` and `patch` have different algorithms for resolving conflicts and applying patches. It's true that `patch` will work in most cases better that `git-am`, but not always.
That's why we have both commands.

-  As for the order: it matters. The main advantage of using `git-am` is debugging. For instance, think of a CVE requiring 10 patches, and the 9th one causing a conflict. The user can go to the worktree and check the conflicting lines and see the current patch context. Whereas with `patch` alone the user would be overwhelmed with the previous 8 patches code in the git workspace. And lastly, fixing the conflicts with `git-am` is just a matter of changing a few lines and `git-am --continue`.

It's true though that the code in klp-build becomes more bloated by using `git-am`, but I think it's worth it. What do you think?

klpbuild/plugins/extract.py
fgyanz: ~~humm I think you are using and outdated version.~~

klpbuild/plugins/extract.py
fgyanz: ~~humm I think you are using and outdated version.~~

vmezzela: mmh It's the last version of the commit "plugins/extract: Apply the patches with git"

klpbuild/plugins/extract.py
vmezzela: Oh, I never seen it that way: when a patch failed to apply I'd usually check if we missed another one rather than adjusting the tree manually.

Does anyone  adjust it manually?

klpbuild/plugins/extract.py
vmezzela: Same as previous comment

klpbuild/klplib/templ.py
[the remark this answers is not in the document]
vmezzela: Hello, turns out this can create issues at runtime: it makes the module notifier return early without solving some externalizations.

Hence we should fix it and use only the name of the module rather than the path.

marcosps:
@fgyanz one thing that ocurred to me when I was rebasing my symbol-validation branch on top of the automation: the `--archs` now have no more usage. Even if we force an arch, but still set a `--conf`, `klp-build` will check the affected archs by using that `CONF`.

We have some tests that are using `--arch x86_64`, let's say, and using a config like `CONFIG_FS_PROC`, but klp-build sets the archs to all of them, since `CONFIG_FS_PROC`. To fix that, if we really want to create a fix for only one arch, in this test, I changed the config to `CONFIG_X86`, just in case.

We don't need to do anything for it right now, it's just a thing that I found, and I believe that we can finally remove the `--archs`, since it's inferred by the config right now.

vmezzela:
Hi,

> We don't need to do anything for it right now, it's just a thing that I found, and I believe that we can finally remove the `--archs`, since it's inferred by the config right now.

I think is better if we keep it. I know that we can change the config, but that's a bit hacky. What do you think?

fgyanz:
I agree with @vmezzela here, for the same reason. '--arch' is more user-friendly than some obscure config value that might  change between kernel versions. 

`We have some tests that are using --arch x86_64, let's say, and using a config like CONFIG_FS_PROC, but klp-build sets the archs to all of them, since CONFIG_FS_PROC. To fix that, if we really want to create a fix for only one arch, in this test, I changed the config to CONFIG_X86, just in case.`

Unless there's a bug, the expected behavior should not be that. If we set the config to  `CONFIG_FS_PROC` and the archs to `x86_64`, the affected codestreams should be within the intersection of `CONFIG_FS_PROC & --archs`. Meaning, all the codestreams that support CONFIG_FS_PROC in x86_64 arch.

marcosps:
> I agree with @vmezzela here, for the same reason. '--arch' is more user-friendly than some obscure config value that might change between kernel versions.
> 
> `We have some tests that are using --arch x86_64, let's say, and using a config like CONFIG_FS_PROC, but klp-build sets the archs to all of them, since CONFIG_FS_PROC. To fix that, if we really want to create a fix for only one arch, in this test, I changed the config to CONFIG_X86, just in case.`
> 
> Unless there's a bug, the expected behavior should not be that. If we set the config to `CONFIG_FS_PROC` and the archs to `x86_64`, the affected codestreams should be within the intersection of `CONFIG_FS_PROC & --archs`. Meaning, all the codestreams that support CONFIG_FS_PROC in x86_64 arch.

Sure, we can look at this later than. Just saying that, in my tests, this wasn't working as it was in the past. But I agree, one thing at a time. This won't hurt having it set like this for now.

klpbuild/plugins/extract.py
fgyanz: My bad, rebase chaos. Fixed

klpbuild/plugins/extract.py
fgyanz: My bad, rebase chaos. Fixed

klpbuild/plugins/extract.py
```diff
@@ -380,32 +380,79 @@ def group_equal_files(lp_name, working_cs):
...
+        bname = get_lp_branch(lp_name, cs)
+        err = subprocess.run(["git", "branch", "-D", f"{bname}"], cwd=sdir,
+                             stdout=subprocess.DEVNULL,
+                             stderr=subprocess.PIPE, check=False)
+        if err.returncode != 0 and f"'{bname}' not found" not in str(err.stderr):
+            raise RuntimeError(f"Failed to delete branch {bname}:\n{err.stderr}\n")
+
+
+def apply_patch(patch, sdir):
+    # Try to apply the patch first with git-am.
```
vmezzela, line 426: Can we have a `logging.info()` here? A bit of time is being spent here and I don't like `klp-build` being totally silent for so long :D

vmezzela:
Note: if I have kgraft-patches currently checked out on one of the livepatch branches, then the `commit` functionality wouldn't fully work. I don't know how this needs to be handled. But it was worth to mention it:

```
╭─vmezzela@localhost ~/Work/kgraft-patches ‹bsc1249537_15.7rtu0-4_15.7u0-4›
╰─$ klp-build commit -n bsc1249537 --force
Commiting livepatch to kgraft...
Failed to delete branch: +bsc1249537_15.7rtu0-4_15.7u0-4: error: branch '+bsc1249537_15.7rtu0-4_15.7u0-4' not found

Livepatch 'bsc1249537_6.0rtu2-12_6.0u2-11_15.6rtu5-14_15.6u6-15' commited
Livepatch 'bsc1249537_15.7rtu0-4_15.7u0-4' commited
Livepatch 'bsc1249537_16.0rtu0_16.0u0' commited
```

fgyanz:
That's true, but the whole point of having the worktree is to not modify the main repository. Should klp-build fail to delete the branch, it's then up to the user to fix it.

> Failed to delete branch: +bsc1249537_15.7rtu0-4_15.7u0-4: error: branch '+bsc1249537_15.7rtu0-4_15.7u0-4' not found

What's up with this error message? Oo and the `+bsc1249537_15.7rtu0-4_15.7u0-4` !!

vmezzela:
> That's true, but the whole point of having the worktree is to not modify the main repository. 

Indeed

> Should klp-build fail to delete the branch, it's then up to the user to fix it.

Agree

> 
> > Failed to delete branch: +bsc1249537_15.7rtu0-4_15.7u0-4: error: branch '+bsc1249537_15.7rtu0-4_15.7u0-4' not found
> 
> What's up with this error message? Oo and the `+bsc1249537_15.7rtu0-4_15.7u0-4` !!

It's the reason why I posted the comment :D 
Should we gracefully exit if that happens? Or at least inform the user what's happening and how to handle it (e.g. checkout to a different branch)

klpbuild/plugins/push.py
```diff
@@ -71,62 +94,29 @@ def create_lp_package(osc, lp_name, i, total, cs):
...
-    if code_path.exists():
-        shutil.rmtree(code_path)
-
     osc.packages.checkout(prj, "klp", prj_path)
 
-    product_branch = cs.get_full_product_name()
-
-    logging.info("(%s/%s) pushing %s using branches %s/%s...",
-                 i, total, cs.full_cs_name(), product_branch, str(branch))
+    push_branch = create_push_branch(cs, branch)
```
vmezzela, line 106: In the future, it might be good to have an option to use as a fallback when this strategy doesn't work because of conflicts. Currently, when it happens I temporarily adjust the code so that it can at least push what I need and then I revert the change.. but that's not optimal.

klpbuild/klplib/kgraft.py
```diff
@@ -0,0 +1,110 @@
...
+def find_lp_branches(pattern):
+    err = subprocess.run(["git", "branch", "--list", pattern],
+                         cwd=get_kgraft(),
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE, text=True,
+                         check=False)
+    if err.stderr or not err.stdout:
+        return None
+
+    return [re.sub(r"[\s\*]",'', l) for l in err.stdout.splitlines()]
```
vmezzela:
```suggestion
    return [re.sub(r"[\s\*+]",'', l) for l in err.stdout.splitlines()]
```


Also this case need to be handled. Otherwise when in the main `kgraft-patches` repository you have one of this branch checked out, this function will return `+bsc12345_*` rather than `bsc12345`.

marcosps (approved):
LGTM!

klpbuild/klplib/kgraft.py
fgyanz: done, thanks a lot!

klpbuild/plugins/extract.py
fgyanz: sure, done!

klpbuild/plugins/push.py
fgyanz: Makes sense. I added the original livepatch branch as a fallback. It should be enough to catch compilation issues, but for nothing else. Not sure if you had something else in mind though. Is there a strategy other than rebasing that we can use as fallback?

klpbuild/plugins/push.py
```diff
@@ -46,13 +53,36 @@ def create_prj_meta(cs):
...
+        create_lp_branch(push_branch, lp_branch)
+        rebase_lp_branch(push_branch, product_branch)
+    except:
+        # Fallback in case the push_branch cannot be successfully created.
+        # Not meant for building final livepatches.
+        logging.warning("Failed to create branch '%s'. Using instead '%s'.",
+                        push_branch, lp_branch)
+        push_branch = lp_branch
+
+    return push_branch
```
vmezzela, line 71: Sorry, what I meant was to have an actual option like `--no-rebase` that skips the rebase. Because it needs to be done only when the user decides so. This of course can't work with codestreams groups but only with one codestream that has been previously manually rebased properly. 

The drawback of having an automatic fallback is that if the rebase fails and the fallback path is taken, most likely it would fail the build, because the user did not rebase the branch in advance manually..

What do you think?

fgyanz: > Sorry, what I meant was to have an actual option like --no-rebase that skips the rebase. Because it needs to be done only when the user decides so. This of course can't work with codestreams groups but only with one codestream that has been previously manually rebased properly.

Ah I see! Yes, it makes sense. But for the future, indeed :D 

> The drawback of having an automatic fallback is that if the rebase fails and the fallback path is taken, most likely it would fail the build, because the user did not rebase the branch in advance manually..

Correct, that's expected. But I see two advantages for having this:
- It will attempt to compile the livepatch files. And although it will never fully succeed, it's good enough for catching some errors (like typical C mistakes). Of course, it's not meant for building livepatches, but it might be good enough during development if for whatever reason the product branches are broken.

- Sometimes only one product branch is conflicting or broken, in such case klp-build will currently crash due to the failed rebased. With this changes, it will at least continue pushing the rest of branches. 

Then again, this change does not fully address the case in your comment. But anyway, it might be somewhat useful :)

vmezzela:
LGTM!

PR #161, opened 2025-10-21
plugins: status: Use termcolor and tabulate libraries

marcosps:
Now it shows a nicer grid and with colored output.

@fgyanz it's using the tabulate lib that you used on klp-bugzilla

### DIFF
--- a/klpbuild/plugins/status.py
+++ b/klpbuild/plugins/status.py
@@ -7,6 +7,8 @@ import logging
 import time
 
 from osctiny import Osc
+from tabulate import tabulate
+from termcolor import colored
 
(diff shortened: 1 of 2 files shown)


klpbuild/plugins/status.py
```diff
@@ -46,16 +58,23 @@ def status(lp_name, lp_filter, wait=False):
 
         logging.info("%d codestreams to finish", len(prjs))
```
vmezzela: What about adding a `os.system('clear')` just before this? or something similar to clean up the terminal before the next iteration.

I am not sure how it works with `logging.info` though.

PR #162, opened 2025-10-22
extract: drop compiler named address space not supported by klp-ccp

vmezzela:
In SLE16 kernels, `CC_HAS_NAMED_AS` config is made available, which enables the use of compiler Named Address Space.

Such feature introduce two new tokens, __seg_gs and __seg_fs, that klp-ccp parser doesn't recognize, hence it fails. Just drop them using the precompiler.

Such issue is only present in sle >= 16, but adding in this branch should suffice and doesn't create issues if the macro is enabled even for some older codestreams.

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -71,8 +71,9 @@ class Codestream:
 
 
     def get_src_dir(self, arch=ARCH, init=True):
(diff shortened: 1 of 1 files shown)


marcosps:
@vmezzela is that works for us, I'm ok with it.

marcosps (approved):
LGTM

PR #163, opened 2025-10-23
Validate sym traceability

marcosps:
Hey guys, this is something that I studies recently, and it would help to detect some corner cases of livepatch creation. The possibillity of having the livepatch procedure to be even more automatic makes such improvement even more important, since we can catch these issues right away.

Please take a look and say what you think about it!


klpbuild/klplib/codestream.py
```diff
@@ -368,12 +369,46 @@ def lp_out_file(self, lp_name, fname):
...
+        start = start_mcount["st_value"] - sec["sh_addr"]
+        stop = stop_mcount["st_value"] - sec["sh_addr"]
+
+        sec_data = sec.data()[start:stop]
+        syms = []
+
+        for i in range(0, len(sec_data), 8):
+            ptr = sec_data[i:i+8]
+
+            data = int.from_bytes(ptr, byteorder="little")
```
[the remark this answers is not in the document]
marcosps: True! This was required when I tried to print the addresses. I'll check if this is necessary for us to adjust depending on the arch, so keep this on hold for me. I'm planning to extend the test to all archs so we can validate this properly.

vmezzela:
Hi Marcos, thanks for the PR! I have one question, as far as I can see with this PR also symbols to be externalized are being checked. Is that intended? Should we check the traceability of the affected symbols only?

marcosps:
> Hi Marcos, thanks for the PR! I have one question, as far as I can see with this PR also symbols to be externalized are being checked. Is that intended? Should we check the traceability of the affected symbols only?

That's true, it was not the intention. I was so focused on the setup that I forgot that this function is also called on extraction. I'll change it the code to only check this on setup phase. Good catch!

vmezzela (changes requested):

marcosps:
> > @vmezzela I believe that now we are covered! And also I removed an argument from another method that seems unnecessary now. Would you mind taking a second look?
> 
> Sure, I added a comment. Sorry, I just spotted it now.
> 
> Thank for dropping that argument, I think it was indeed unnecessary :)

Well, it seems that it was not. We need to check for the archs that are affected by the CVE, and the current loop over cs.archs is related to the archs supported on the codestream... Maybe we can add a comment to clarify that :)

marcosps:
The new commits in this branch are necessary for those cases where the livepatch doesn't affect x86, and so setup or extract could fail, since we always check things against x86.

marcosps:
@fgyanz this is now ready, would you mind reviewing it again? It's passing all tests.

klpbuild/plugins/extract.py
```diff
@@ -806,7 +806,7 @@ def start_extract(lp_name, lp_filter, no_patches, avoid_ext):
                 obj_syms[obj].extend(syms)
 
         for obj, syms in obj_syms.items():
-            missing = cs.check_symbol_archs(get_codestreams_data('archs'), obj, syms, True)
+            missing = cs.check_symbol_archs(get_codestreams_data('archs'), obj, syms, True, False)
             if missing:
```
fgyanz: I think it's safe to remove this guard. The function returns {}, which can .items().

klpbuild/klplib/codestream.py
```diff
@@ -382,24 +384,100 @@ def lp_out_file(self, lp_name, fname):
...
             syms = symtab.get_symbol_by_name(symbol)
             # can return None is the symbol is not found, or a list if the symbol
             # is not unique
             if not syms:
                 ret.append(symbol)
+                continue
 
-            elif len(syms) > 1:
+            if len(syms) > 1:
                 print(f"WARNING: {self.full_cs_name()}-{arch} ({self.kernel}): symbol {symbol} duplicated on {mod}")
```
fgyanz: Do you mind mind changing this print to logging.warning() ?

klpbuild/klplib/codestream.py
```diff
@@ -382,24 +384,100 @@ def lp_out_file(self, lp_name, fname):
         fpath = f'{str(fname).replace("/", "_").replace("-", "_")}'
         return f"{lp_name}_{fpath}"
 
+    def __check_patchable_sym(self, arch, name, sym, trace_addrs):
+        val = sym["st_value"]
+
+        pos = bisect.bisect_left(trace_addrs, val)
+        if pos == len(trace_addrs) or trace_addrs[pos] != val:
+            raise RuntimeError(f"{self.full_cs_name()}-{arch} ({self.kernel}): Symbol {name} has tracing disabled.")
```
fgyanz: Would it be possible to use logging.error and exit() instead of RuntimeError? The latter is too verbose.
Same for the rest of RuntimeErrors

klpbuild/plugins/extract.py
```diff
@@ -270,6 +270,7 @@ def get_klpp_symbols(out_dir, lp_out):
                 logging.warning(f"Failed to find klpp_{sym} in {lp_out}")
                 continue
             klpp_proto = re.sub(r'\s+',' ', m.group(2)).strip() + ';'
+            klpp_proto = re.sub(r' __(init|exit)', ' ', klpp_proto)
```
fgyanz, line 275: Do you mind adding a simple comment clarifying why this is needed? Same below :)

fgyanz (commented):
Hi @marcosps , the PR looks great! Just some minor nitpicking here and there.

One last thing, would it be possible to raise a big `WARNING` at the end of `extract` command in case a user decides to generate livepatches for a non-x86_64 arch? 
The reason being that `klp-ccp`  does not currently support other archs, so generated livepatches won't probably work. Even worse, they might compile and run, but underneath the code might be wrong. I think we have the responsibility of alerting the user in such case :)

Thanks a lot for the good work :partying_face:

klpbuild/klplib/file2config.py
```diff
@@ -97,6 +97,17 @@ def find_configs_for_files(cs, file_paths: list):
...
+
+        # Detect code that is only enabled on a specific architecture
+        if path.startswith("arch"):
+            if "s390" in path:
+                configs[path] = {'conf': "CONFIG_S390", 'module': 'vmlinux'}
+            elif "x86" in path:
+                configs[path] = {'conf': "CONFIG_X86_64", 'module': 'vmlinux'}
+            elif "powerpc" in path:
+                configs[path] = {'conf': "CONFIG_PPC64", 'module': 'vmlinux'}
+            continue
```
fgyanz: I think we can make it a bit more robust by doing it just after the `_find_config()` and:
```suggestion
        if path.startswith("arch") and len(cs.get_all_configs(conf)) > 1:
            if "s390" in path:
                configs[path] = {'conf': "CONFIG_S390", 'module': 'vmlinux'}
            elif "x86" in path:
                configs[path] = {'conf': "CONFIG_X86_64", 'module': 'vmlinux'}
            elif "powerpc" in path:
                configs[path] = {'conf': "CONFIG_PPC64", 'module': 'vmlinux'}
            continue
```
So if the found `conf` is too generic, fallback to the arch specific one. What do you think? :)

klpbuild/plugins/extract.py
marcosps: Done

klpbuild/klplib/codestream.py
marcosps: done

klpbuild/klplib/codestream.py
marcosps: done

klpbuild/plugins/extract.py
marcosps: done

klpbuild/klplib/file2config.py
marcosps: done

marcosps:
@fgyanz ready for another revision :)

fgyanz (approved):
Thanks a lot for the changes! LGTM!

PR #174, opened 2025-11-21
Add Format patches '--no tests'

fgyanz:
Add an option for exporting patches without tests.
This will avoid triggering warnings about missing tests and some other corner cases.


marcosps (approved):
Great, thanks!

PR #176, opened 2025-11-24
klplib/ksrc: Blacklist kmp modules

fgyanz:
kmp modules are not currently supported by klp-build. If found, trigger a warning to let the user know.

### DIFF
--- a/klpbuild/klplib/patch.py
+++ b/klpbuild/klplib/patch.py
@@ -5,6 +5,7 @@
 
 import logging
 
+from pathlib import PurePosixPath
 from collections import defaultdict
 
 from klpbuild.klplib import utils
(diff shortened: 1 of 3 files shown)


klpbuild/klplib/patch.py
```diff
@@ -162,7 +163,11 @@ def analyse_kmodules(cs_list):
                            cs.configs[conf].items()]:
                 continue
 
-            supported = cs.is_module_supported(mod)
+            supported, blacklisted = cs.is_module_supported(mod)
+            if blacklisted:
+                logging.warning("%s: Module '%s' is not supported by klp-build.",
+                                cs.full_cs_name(), os.path.basename(mod))
```
marcosps: To keep using pathlib, we can use PurePath.name here (as described in https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name), and drop the `os` one.

marcosps (approved):
Good work! Just one small nit to make things use the same libs that we are currently using. Once that one is fixed I believe that this can be merged.

klpbuild/klplib/patch.py
fgyanz: Thanks! Done :)

PR #178, opened 2025-11-26
New  kgr test path

fgyanz:
Some machines, although not all, have a new 'klp' user which will be the new fallback path for vms data. The end goal is to get rid of other users vms and have a shared directory to save space.


marcosps (approved):
LGTM!

PR #183, opened 2025-12-09
Fix architecture check on TemplateGen against the archs supported by the codestream

marcosps:
The original check was working by change. Each codestream has a set of supported architectures, like -RT ones being supported only on x86_64 for now. The current check was checking the supported archs on a codestream with the architectures that was affected on the project level, which is wrong.

The change checks the affected architectures on the codestream against the architectures that the codestream supports. The original tests also worked with this change, being even more clear that it worked by chance at the time.


marcosps:
@fgyanz @vmezzela I just added a new commit to help reducing the time spent on setup. The next improvement should be by parallelizing it by by the same number of jobs that we currently use for extraction. I can think about it in the next days...

@fgyanz would you mind trying this PR on CI to check if this helps to reduce the slowness that you found there?

klpbuild/plugins/setup.py
```diff
@@ -191,7 +193,19 @@ def setup_project_files(lp_name, codestreams):
             for arch in archs:
                 mod = cs.get_file_mod(f, arch)
                 __setup_check_mod(cs, mod, arch)
-                __setup_check_syms(cs, mod, syms, arch)
+
+                slist.setdefault(mod, {})
+
+            data = slist.get(mod, {})
```
fgyanz: Just nitpicking, so feel free to skip this comment. 
I'd propose renaming this variable to something more specific. "data" could mean anything in this context. I'd also propose changing "slist" to "syms" if it's a list of symbols. I think it would be more readable that way.

.github/workflows/ci-tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - name: Run pylint
       run: >
         tox -e lint --
-        --fail-under 9.85
+        --fail-under 9.83
```
fgyanz, line 30: I'm not really sure this is the proper fix. I mean, it works as a workaround now, but will we need to keep doing it every now and then? Maybe we should rethink the approach for a more solid solution.

fgyanz (approved):
Just two comments! Not mandatory to fix them, feel free to decide what to do. Thanks for working on the PR so fast! LGTM!

.github/workflows/ci-tests.yml
marcosps: Yeah, but unfortunately pylint shows really shows what decreased last time, and when I try to run pylint locally it still shows the score as 9.86, so I'm not sure that changed...

klpbuild/plugins/setup.py
marcosps: There is already a variable called `syms` in the same function =/

klpbuild/plugins/setup.py
```diff
@@ -244,6 +244,13 @@ def __setup_check_mod(cs, mod, arch):
 def __setup_check_syms(cs, mod, syms, arch):
     # Verify if the functions exist in the specified object
     arch_syms = cs.check_symbol_archs(arch, mod, syms, False, True)
+
+    # We don't need to check for missing symbols on the HOST (x86_64?) since some
+    # symbols are inlined and klp-ccp will correctly address it. But it might be
+    # interesting to check in other architectures.
+    if arch == utils.ARCH:
+        return
+
```
vmezzela: I am not completely sure this is the right way to go. If a symbol is not present in `x86_64` and on the other archs, then klp-build would still log the warning for the other archs even though the code extracted by klp-ccp would (hopefully) be fine assuming the inlining decision is the same across all the archs. Hence, we'd have the warning in a totally useless case.

Consider the scenario where a given symbol is inlined only in `x86_64` and not on the other archs, there is no way for us to find out that on the other archs we need to patch those syms as well. Hence, here we'd miss the warning in a totally useful case.

So, rather that dropping the warning altogether for `x86_64`, I'd propose to log it anyway. But amending the message so that the user knows that it's not necessarily an issue in case the symbol is missing.

klpbuild/plugins/setup.py
```diff
@@ -191,7 +193,19 @@ def setup_project_files(lp_name, codestreams):
             for arch in archs:
                 mod = cs.get_file_mod(f, arch)
                 __setup_check_mod(cs, mod, arch)
-                __setup_check_syms(cs, mod, syms, arch)
+
+                slist.setdefault(mod, {})
```
vmezzela: I think that this line can be remove altogether: it's redundant with the `slist.get(mod, {})` ?

klpbuild/plugins/setup.py
vmezzela: I agree, I would change the name of the dict and make it clear what is it meant for. Something like `syms_to_be_checked`.

vmezzela (changes requested):

klpbuild/plugins/setup.py
fgyanz: > I am not completely sure this is the right way to go. If a symbol is not present in x86_64 and on the other archs, then klp-build would still log the warning for the other archs even though the code extracted by klp-ccp would (hopefully) be fine assuming the inlining decision is the same across all the archs. Hence, we'd have the warning in a totally useless case.

right, good catch. I'd avoid printing the warning if all the archs are missing the symbol. Most of the times this is case, which means the symbols is inlined.

marcosps: Ok, let me change the patch then!

klpbuild/plugins/setup.py
marcosps: I believe so. I changed this code so many times that somehow it was needed at some point...

klpbuild/plugins/setup.py
marcosps: agreed!

marcosps:
@fgyanz @vmezzela so I changed the check for showing the warning for missing symbols, please take a look!

fgyanz:
Changes make sense! LGTM!

klpbuild/klplib/codestream.py
```diff
@@ -194,6 +194,7 @@ def set_configs(self, configs):
 
     def set_archs(self, archs):
         self.archs = list(set(archs) & set(self.get_default_archs()))
+        self.archs.sort()
```
vmezzela: This commit made me realize that `__is_check_enabled()` relies on the two lists  `cs.archs != cs.get_default_archs()` being the same also in their order, as you mentioned in the commit message.

If that's the case, can we readapt the code to use a `set()` for the archs field in codestreams?

So we avoid similar bugs in the future around this code. Feel free also to add a patch on top of the current branch instead of rebasing - otherwise I can also address this in my PR.

klpbuild/plugins/setup.py
```diff
@@ -242,9 +241,40 @@ def __setup_check_mod(cs, mod, arch):
...
+        mod_syms.update(cs.check_symbol_archs(arch, mod, syms, False, True))
+
+    # Reverse the dict, checking in which architectures are the symbol is not
+    # present. It it's not found on all architectures, it can mean that it's
+    # inline in all of them, so it's mostly harmless, since klp-ccp can handle
+    # it.
+    sym_archs = {}
+    for larch, lsyms in mod_syms.items():
+        for sym in lsyms:
+            sym_archs.setdefault(sym, [])
```
vmezzela, line 265: If you decide to apply the change suggested in the previous comment, this would became `sym_archs.setdefault(sym, set())`, and you can drop the `larchs.sort()` below so we no longer have to rely on the list being sorted (we might forget about that in the future).

vmezzela (changes requested):
Hi @marcosps, the PR looks good to me. I left a proposed change anyway if you want to address it. I believe it'd make the code less prone to future bugs around the way we use `cs.archs`. Otherwise I apply it in my PR if you like the proposal.

marcosps:
> Hi @marcosps, the PR looks good to me. I left a proposed change anyway if you want to address it. I believe it'd make the code less prone to future bugs around the way we use `cs.archs`. Otherwise I apply it in my PR if you like the proposal.

Done! Thanks for the suggestion! Adapting the code to a set required some more changes, but in the end it was worth it!

Also, in the symbol check, I had to keep it it as a list, and then later on convert it as a set, otherwise we can't append the archs into set. I believe that I addressed all your comments!

PR #187, opened 2025-12-17
plugins/extract: Apply the patches in numerical order

fgyanz:


marcosps:
Fixed my issue, thanks!

PR #188, opened 2025-12-18
extract: fix cleanup regexp

vmezzela:
Use `^\w` to make sure the match begins with a character.

Then replace `.*` with `[^;:()]` to have a hard guard against label and to make sure we don't overflow after the `;`.

This is just an initial version. If you have ideas to make it more strict I'd appreciate it, I'd rather having this more strict and therefore get some false negative than matching false positive :)

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -264,7 +264,7 @@ def fix_ext_symbols(cs, lp_dat, lp_out):
     for sym_list in lp_dat["ext_symbols"].values():
         for s in sym_list:
             # Keep only static declarations
-            lp_out = re.sub(rf"^(?!static|\s|#).*\s+\(\*klpe_{s}\)[\s\S]*?;",
+            lp_out = re.sub(rf"^(?!static|\s|#)\w[^;:()=]+\(\*klpe_{s}\)[^;:=]*;",
                             '', lp_out, flags=re.MULTILINE)
 
 


marcosps (approved):
LGTM, thanks a lot!

Maybe we can add a test for this case specifically? Since the bug was found on all 15.3 codestreams, maybe we can just check if the function remains there.

Either way, since solved a bug, so we can merge it right now if you prefer.

fgyanz:
Thanks for the PR! Just two ideas:
- Maybe we should guard against assignments such as: `demo = (*klpe_test);` I guess this could be found in expanded macros.
- Instead of entirely removing the klpe functions declaration, perhaps we should just comment it out. That would allow to easily detect any mistakes.

vmezzela:
>     * Maybe we should guard against assignments such as: `demo = (*klpe_test);` I guess this could be found in expanded macros.

Right, I will address it.

> 
>     * Instead of entirely removing the klpe functions declaration, perhaps we should just comment it out. That would allow to easily detect any mistakes.

I am not sure that would work. As you said, if the extracted source is too big, there's no chance we can catch such things and very likely it would go unnoticed if it compiles and run.

Opting for this change would only give us the possibility to check once we realized that something wrong happened, because we can see the commented-out code. But we would achieve the same just be leaving the `work_*.c` file unchanged and cleaning up directly on the final lp file and running `diff` on them.

But still, even after implementing it, we'd need some **proactive** strategy to catch cases like this.

fgyanz (approved):
humm yes, but grepping for comments during the reviewing would probably be enough. On the other hand, yes, ideally the better solution should be leaving the` work_*.c` file unchanged and then diffing.

Anyway, whatever we decide to do, it's probably outside the scope of this PR. 
Thanks for the changes!

Issue #186, opened 2025-12-12
setup: Invalid config for arch/s390/kernel/perf_cpum_sf.c


fgyanz:
The setup is reporting that `arch/s390/kernel/perf_cpum_sf.c` requires `CONFIG_PERF_EVENTS` in` 6.0rtu4 15.6rtu5-6`, which is not true. The reported kernel modules is correct though.

```
[2025-12-11 19:24:14] 6.0rtu4 15.6rtu5-6:
[2025-12-11 19:24:14] FILE: arch/s390/kernel/perf_cpum_sf.c
[2025-12-11 19:24:14] CONF: CONFIG_PERF_EVENTS
[2025-12-11 19:24:14] OBJ: arch/s390/kernel/perf_cpum_sf
[2025-12-11 19:24:14] FUNCS: cpumsf_pmu_stop
[2025-12-11 19:24:14] 
[2025-12-11 19:24:14] 6.0u4 12.5u62-63 15.3u50-52 15.4u33-35 15.5u21-22 15.6u6:
[2025-12-11 19:24:14] FILE: arch/s390/kernel/perf_cpum_sf.c
[2025-12-11 19:24:14] CONF: CONFIG_S390
[2025-12-11 19:24:14] OBJ: vmlinux
[2025-12-11 19:24:14] FUNCS: cpumsf_pmu_stop
[2025-12-11 19:24:14] 
[2025-12-11 19:24:14] [*] Analysing required CONFIGs...
[2025-12-11 19:24:14] 
[2025-12-11 19:24:15] 6.0rtu4 15.6rtu5-6:
[2025-12-11 19:24:15] CONF: CONFIG_PERF_EVENTS
[2025-12-11 19:24:15] x86_64: y
[2025-12-11 19:24:15] ppc64le: n
[2025-12-11 19:24:15] s390x: n
[2025-12-11 19:24:15] 
[2025-12-11 19:24:15] 6.0u4 12.5u62-63 15.3u50-52 15.4u33-35 15.5u21-22 15.6u6:
[2025-12-11 19:24:15] CONF: CONFIG_S390
[2025-12-11 19:24:15] x86_64: n
[2025-12-11 19:24:15] ppc64le: n
[2025-12-11 19:24:15] s390x: y
[2025-12-11 19:24:15] ...
[...]

PR #190, opened 2025-12-22
Fix arch bugs

fgyanz:
- Slighly rework the arch heuristic so that it returns a
per-architecture generic CONFIG only if the found CONFIG
does not affect the same architecture as the one indicated
in the given file path. (Fixes: #186)

- Missing symbol warning:
The warning should be raised only if the symbol is
missing in a subset of the architectures where the
CONFIG is enabled for a specific codestream.

### DIFF
--- a/klpbuild/klplib/file2config.py
+++ b/klpbuild/klplib/file2config.py
@@ -9,6 +9,14 @@ import re
 
 from pathlib import Path
 
+
+archs_config = {
+        's390x': {'conf': "CONFIG_S390", 'module': 'vmlinux'},
+        'x86_64': {'conf': "CONFIG_X86_64", 'module': 'vmlinux'},
+        'ppc64le': {'conf': "CONFIG_PPC64", 'module': 'vmlinux'},
+}
+
+
 BLACKLIST = {
         r"drivers/gpu/drm/amd/(?!amdgpu/).*":
         "drivers/gpu/drm/amd/amdgpu/amdgpu_irq.c"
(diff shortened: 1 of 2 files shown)


marcosps (approved):
Lgtm!

klpbuild/klplib/file2config.py
```diff
@@ -103,14 +122,15 @@ def find_configs_for_files(cs, file_paths: list):
...
-                configs[path] = {'conf': "CONFIG_X86_64", 'module': 'vmlinux'}
-            elif "powerpc" in path:
-                configs[path] = {'conf': "CONFIG_PPC64", 'module': 'vmlinux'}
+        # Detect code that is only enabled on a specific architecture.
+        # Use a per-architecture generic CONFIG only if the found CONFIG
+        # does not affect the same architecture as the one indicated in
+        # the given file path.
+        elif path.startswith("arch"):
+            archs = cs.get_all_configs(config)
+            arch = _get_arch_in_path(path)
```
vmezzela, line 131: Given that `__get_arch_in_path(path)` is called within the assumption that `path.startswith("arch")`, I would put an `assert(arch)`. Otherwise, it would be really weird to let this run even if it returns `None` in the case the path was on another unsupported arch.

vmezzela (changes requested):
Just a small nit, otherwise lgtm!

klpbuild/klplib/file2config.py
fgyanz: After further discussions we agreed on keeping it as is.

PR #191, opened 2025-12-31
push: always build all archs

vmezzela:
We should build all archs even when the livepatch does not affect one of them to make sure that conditional compilation works.


marcosps (approved):
We have been changing the archs for Codestreams lately that things are still kind of broken in some places... thanks a lot for fixing it!

LGTM!

vmezzela:
Hi Marcos, before we merge, I will push another commit still related to this PR as I just realized we have the same issue during prepare-tests.

marcosps (approved):
That's good, thanks a lot for the fixes!

PR #192, opened 2026-01-02
klplib: codestream: Handle IBT enabled function offsets

marcosps:
When checking if a symbol is traceable, the trace addresses can be 4 bytes ahead of the symbol table. This happens when the symbol has an additional ENDBR instruction, which shifts the __fentry__ by 4 bytes.

In order to correctly identify if a symbol is traceable or not, check both addresses: the one found symbol, and one by shifting the addr by 4 bytes.


marcosps:
@vmezzela would you mind trying this PR and check if this solves your issue? Thanks a lot!

vmezzela (approved):
lgtm!

PR #193, opened 2026-01-16
Push fast

marcosps:
Here I added --fast, inspired by the idea from Fernando for klp-test, but now for pushing. Maybe we can start adding the same flag for another plugins, but it's something for another day.

Plus this new argument, I also added some cleanups in the code around the functionality, and added a new test.

### DIFF
--- a/klp-build.1
+++ b/klp-build.1
@@ -247,6 +247,10 @@ can automate the building phase for each codestream and architecture.
 .TP
 .B --wait
 Wait until all codestreams builds are finished.
+.TP
+.B --fast
+Filter only one codestream from each product. It makes testing a livepatch easier
+than appying a random filter to select one a few codestreams.
 .RE
 .TP
 .B status
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -80,6 +80,9 @@ class Codestream:
                 self.update == cs.update and \
                 self.rt == cs.rt
 
+    def __str__(self):
+        return self.full_cs_name()
+
     def get_src_dir(self, arch=None, init=True):
         # Before sle16, only -rt codestreams have a suffix for source directory
         has_rt_suffix = self.rt and self.sle < 16
--- a/klpbuild/klplib/utils.py
+++ b/klpbuild/klplib/utils.py
@@ -267,6 +267,28 @@ def check_module_unsupported(arch, mod_path):
     return "no" == get_elf_modinfo_entry(elffile, "supported")
 
 
+def filter_fast(cs_list):
+    '''
+    Return the first codestream of each product. For example, given the list
+        15.5u10, 15.5u11, 12.5u10, 6.0u0
+    The return would be
+        15.5u10, 12.5u10, 6.0u0
+
+    Returns:
+        List: Containing only one entry per product
+    '''
+    ret = []
+    cs_once = []
+
+    for c in cs_list:
+        base = c.base_cs_name()
+        if base not in cs_once:
+            ret.append(c)
+            cs_once.append(base)
+
+    return ret
+
+
 def filter_codestreams(lp_filter, cs_list, verbose=False):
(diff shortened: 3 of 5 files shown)


vmezzela (approved):
Thanks a lot! This is really useful!

fgyanz:
Very nice! Just one additional idea: If `filter` is not specified, could we apply `fast` on the codestream groups from `ccp/groups`? 
E.g:
```
# ccp/groups
15.6u5-10 15.6rtu8 6.0u8-9
12.5u50-53 15.4u8-10
```

run: ```klp-build push -n bscxxxx --fast```

And it pushes:

```15.6u5 15.6rtu8 6.0u8 12.5u50 15.4u8```

This is actually being done on our CI, and perhaps it could be useful to the users. What do you think?

**Comment by marcosps** [comment]
> Very nice! Just one additional idea: If `filter` is not specified, could we apply `fast` on the codestream groups from `ccp/groups`? E.g:
> 
> ```
> # ccp/groups
> 15.6u5-10 15.6rtu8 6.0u8-9
> 12.5u50-53 15.4u8-10
> ```
> 
> run: `klp-build push -n bscxxxx --fast`
> 
> And it pushes:
> 
> `15.6u5 15.6rtu8 6.0u8 12.5u50 15.4u8`
> 
> This is actually being done on our CI, and perhaps it could be useful to the users. What do you think?

This is what this PR is about, it works exactly like you described :)

**Comment by fgyanz** [comment]
I might be missing something, but codestream groups are not identical to the codestream list.
For instance:
```
# ccp/groups
15.5u12
15.5u13
```
Those are different groups, and the `fast` I'm proposing would push both, because they are different.
Meanwhile, the current `fast` would `push` only `15.5u12`, right?

PS: It's just an idea, the current PR looks good enough :)

**Comment by marcosps** [comment]
> I might be missing something, but codestream groups are not identical to the codestream list. For instance:
> 
> ```
> # ccp/groups
> 15.5u12
> 15.5u13
> ```
> 
> Those are different groups, and the `fast` I'm proposing would push both, because they are different. Meanwhile, the current `fast` would `push` only `15.5u12`, right?
> 
> PS: It's just an idea, the current PR looks good enough :)

Ah ok. Yeah, you are right. But, at the same time, we can't really just push groups because sometimes very different products will be grouped together, but should be tested separately, like SLE12 (kgraft) and SLE16 (klp-convert).

Maybe we can put more effort into checking this as well when pushing for fast, like first checking groups, but then making sure that we also push different base products? Do you think it's worth the work? I mean, the `--fast` would be used to check quickly if the code compiles...

I can see value in doing and not doing it, just asking what do you think :)

**Comment by fgyanz** [comment]
> But, at the same time, we can't really just push groups because sometimes very different products will be grouped together, but should be tested separately, like SLE12 (kgraft) and SLE16 (klp-convert).

In theory, afaik the proposal should handle that case too. By that I mean:
```
# ccp/groups
12.5u12 16.0u2-4
15.5u12
15.5u13
```
The CI would push: `12.5u12 16.0u2 15.5u12 15.5u13`
So effectively skipping only: `16.0u3 and16.0u4` which share the same code as `15.6u2`.

```
# The CI code doing this part:
filter_cs()
{
     < "$1/ccp/groups" xargs | sed -E 's/-[0-9]{1,2}//g' | tr ' ' '|'
}
```

> 
> Maybe we can put more effort into checking this as well when pushing for fast, like first checking groups, but then making sure that we also push different base products? Do you think it's worth the work? I mean, the --fast would be used to check quickly if the code compiles...
> 
> I can see value in doing and not doing it, just asking what do you think :)

We already have `get_lp_groups()`, but indeed, this will require some checks. I don't think it's worth the effort if it will take more than 5-10mins to implement. Using `ccp/groups` is just a neat feature, but your `fast` already does a good enough job imo.

fgyanz (approved):

PR #194, opened 2026-01-23
setup.py: Add --full-checks argument and disable ftrace checks by def…

marcosps:
…ault

This change is important since it takes a lot of time check if the symbols about to be patched are traceable or not. The common sense here would be to only enable these checks on the CI for now.

For a given livepatch that affects 50+ codestreams, this reduces the time of setup command from 18 minutes to 13 minutes.

### DIFF
--- a/klpbuild/plugins/setup.py
+++ b/klpbuild/plugins/setup.py
@@ -32,6 +32,12 @@ def register_argparser(subparser):
         help="Do not check for already patched codestreams, do the setup for all non filtered codestreams.",
     )
     args.add_argument(
+        "--full-checks",
+        action="store_true",
+        help="Enable advanced checks that verify if symbol can be livepatched. "
+        "Enabling this option can incur in severe slowdowns.",
+    )
+    args.add_argument(
         "--file-funcs",
         required=False,
         action="append",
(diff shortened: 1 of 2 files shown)


klpbuild/plugins/setup.py
```diff
@@ -31,6 +31,11 @@ def register_argparser(subparser):
         action="store_true",
         help="Do not check for already patched codestreams, do the setup for all non filtered codestreams.",
     )
+    args.add_argument(
+        "--full-checks",
+        action="store_true",
+        help="Enables checks that verify if symbol is traceable, and others.",
```
fgyanz: I'd rewrite this a bit. I'd not mention here symbol traceability, because the user might not know what it means. How about:
"Enable adavance checks that verify if the symbol can be livepatched"
Or something on that line.

I'd leave the technical details to the manpage doc. Which reminds me, please add a section in the manpage for this new option :D 

Edit: Also, do you mind mentioning that enabling this option has the side-effect of a big slowdown?

fgyanz (changes requested):
Thanks a lot! Looks good to me. I added a commen with just one small nitpick and a reminder.

vmezzela (approved):
Thanks a lot!
I agree with the comment of Fernando, besides that, lgtm!

fgyanz (approved):
LGTM! Thanks a lot for including the suggested changes :)

PR #195, opened 2026-02-09
Dropping out-of-support codestreams

marcosps:
Currently 15.3 is out of support, and so cve-5.3 branch as well. Also, The only SLE 15.X RT branches supported is SLE15-SP7, so drop all others as well.


vmezzela (approved):
lgtm!

fgyanz:
Hi @marcosps , I think we need to also fix some tests that failing due to missing codestreams (which are now OOS)

marcosps:
@fgyanz fixed everything (tm)!, I'm not sure why the commit that introduced the full_checks arg is now also being added here...

fgyanz (approved):
LGTM! Thanks

PR #196, opened 2026-02-11
Fix missing ifdefs when only one architecture is affected

marcosps:
For one livepatch I found that there were missing ifdefs for an ACPI file, which is weird because only x86_64 has CONFIG_ACPI set. This now makes sure that we add ifdefs if the CONFIG_ entries are not the same as cs.get_default_archs().

### DIFF
--- a/klpbuild/plugins/setup.py
+++ b/klpbuild/plugins/setup.py
@@ -144,6 +144,10 @@ def setup_archs(codestreams):
     archs = utils.affected_archs(codestreams)
     set_codestreams_data(archs=archs)
 
+    # Inspect which codestreams are enabled for the affected architectures
+    for cs in codestreams:
+        cs.archs = (cs.archs & set(utils.affected_archs([cs])))
+
 
 def setup_codestreams(lp_name, data):
     if not lp_name.startswith("bsc"):
--- a/tests/test_templ.py
+++ b/tests/test_templ.py
@@ -138,6 +138,36 @@ def test_is_enabled_only_on_cs_arcs():
         assert "#if IS_ENABLED" not in content
 
 
+# Check if only x86_64 is affected by the CVE, meaning that IS_ENABLED should be
+# set in the final code
+def test_is_enabled_only_on_x86():
+    lp = "bsc_" + inspect.currentframe().f_code.co_name
+    cs = "15.7u5"
+
+    setup_args = {
+        "lp_name": lp,
+        "lp_filter": cs,
+        "no_check": True,
+        "archs": utils.ARCHS,
+        "cve": None,
+        "conf": "CONFIG_ACPI",
+        "module": None,
+        "file_funcs": [
+            ["drivers/acpi/acpica/utcopy.c", "acpi_ut_copy_ipackage_to_ipackage"]
+        ],
+        "mod_file_funcs": [],
+        "conf_mod_file_funcs": [],
+        "full_checks": False,
+    }
+    setup(**setup_args)
+
+    extract(lp_name=lp, lp_filter=cs, no_patches=True, avoid_ext=[])
+
+    content = get_file_content(lp, cs, f"livepatch_{lp}.c")
+    # This CVE impact only x86, so IS_ENABLED should be present
+    assert "#if IS_ENABLED" in content
+
+
 # For multifile patches, a third file will be generated and called
 # livepatch_XXX, and alongside this file the other files will have the prefix
 # bscXXXXXXX.


fgyanz (approved):
LGTM! Tested with a different livepatch too.

klpbuild/klplib/utils.py
```diff
@@ -338,7 +338,7 @@ def affected_archs(cs_list):
         for val in cs.configs.values():
             conf_archs.update(val)
 
-    return sorted(conf_archs)
+    return set(sorted(conf_archs))
```
vmezzela: `conf_archs` is already a set before we call `sorted()` on it, which transforms it into a sorted list. Instead of converting it back to a `set()`, we'd have the same result if we return straight away `conf_arch`.

or am I missing something?

klpbuild/klplib/templ.py
```diff
@@ -637,7 +644,7 @@ def generate_livepatches(lp_name, cs):
 def __is_check_enabled(cs: Codestream):
     # Require the IS_ENABLED ifdef guard whenever we have a livepatch that
     # is not enabled on all architectures of the given Codestream
-    return cs.archs != cs.get_default_archs()
+    return affected_archs([cs]) != cs.get_default_archs()
```
vmezzela: I am not sure I understand why this fix the issue.. is `cs.archs` value not correct? if so, shouldn't we make sure we set it during the setup?

Perhaps a missing `cs.set_archs(archs)` during setup phase? I noticed that it's only called during `setup_manual()`, but not when we let klp-build find the config automatically.

I think that even if this hunk solves the issue, we would still leave a wrong `cs.archs` and we might happen to use it somewhere else with an incorrect value.

Besides that, the old code was comparing to `!= cs.get_default_archs()`, would it make more sense to do `!= ARCHS`  instead? In this way, the ifdef would be emitted in more codestreams, allowing for better grouping.

klpbuild/klplib/utils.py
marcosps: In my tests, it created a set already sorted in this case. Without the set created here, it fails to compare the set's in the templ code.

klpbuild/klplib/templ.py
marcosps: `cs.archs` is incorrectly. It changes so much in the recent times that I'm not even sure it's needed anymore, given that now we have the `configs.CONF_X.$ARCH` saying which architectures have the given CONF_ needed by the given file (the `file.c.conf` can be used to check explicity if the given CONF is enabled or not).

I believe so, but even more important is that, right now we only set the archs at the start of the setup process, and later on we find out which archs are really impacted.

We can't use `ARCH` to compare because it uses all currently supported ARCHS, but we need to check which archs are supported for the given codestream (MICRO don't support ppc64le IIRC).

About the wrong cs.archs, I agree, but it seems much easier to rely on the `configs` now that I believe that we could rip it off. What do you think?

The current change fixed my issue and passed the tests, so for now it seems correct (I hope so).

vmezzela: > I believe so, but even more important is that, right now we only set the archs at the start of the setup process, and later on we find out which archs are really impacted.

Exactly. What I meant in the previous comment is: should we update `cs.archs` after we find out with archs are really impacted?

> We can't use `ARCH` to compare because it uses all currently supported ARCHS, but we need to check which archs are supported for the given codestream (MICRO don't support ppc64le IIRC).

Using `cs.get_default_archs()` is correct. However, using `ARCHS` will lead to more codestreams grouped together.

If you have a CVE that affects only `x86_64`, then the `#ifdef` will be emitted on `-default` codestreams (`cs.archs != cs.get_default_archs()`) but not on `-rt`  (in this case `cs.archs == cs.get_default_archs()`). Therefore the different `#ifdef` emission decision across codestreams will split the grouping even more. I hope I clarified my point.

It's not mandatory, I just think it's an improvement while on it. But feel free to discard this suggestion :P
 
> About the wrong cs.archs, I agree, but it seems much easier to rely on the `configs` now that I believe that we could rip it off. What do you think?

I suspect we still use `cs.archs` somewhere else. If so, isn't ripping it off much of a bigger job than just updating it to a correct value during setup?

> 
> The current change fixed my issue and passed the tests, so for now it seems correct (I hope so).

Sure thing. I understand that it fixes the `#ifdef` emission issue. What I wanted to bring up is that if `cs.archs` is wrong, as you confirmed, then there might be a problem wherever we are still using such field.

marcosps: > > I believe so, but even more important is that, right now we only set the archs at the start of the setup process, and later on we find out which archs are really impacted.
> 
> Exactly. What I meant in the previous comment is: should we update `cs.archs` after we find out with archs are really impacted?

I believe so, but it would carry a big amount of changes, and in the end we might find out that we don't need `cs.archs` anymore :)

> 
> > We can't use `ARCH` to compare because it uses all currently supported ARCHS, but we need to check which archs are supported for the given codestream (MICRO don't support ppc64le IIRC).
> 
> Using `cs.get_default_archs()` is correct. However, using `ARCHS` will lead to more codestreams grouped together.
> 
> If you have a CVE that affects only `x86_64`, then the `#ifdef` will be emitted on `-default` codestreams (`cs.archs != cs.get_default_archs()`) but not on `-rt` (in this case `cs.archs == cs.get_default_archs()`). Therefore the different `#ifdef` emission decision across codestreams will split the grouping even more. I hope I clarified my point.
> 

In this case, I believe that `cs.archs` in wrong in both, at least on my current livepatch. When using the automatic scan, in my case at least, `cs.archs` is set to `ARCHS`, so it's always all on them in this case. That's why it's missing the `ifdef` in my case, since it thinks that it enabled on all architectures.

> It's not mandatory, I just think it's an improvement while on it. But feel free to discard this suggestion :P

Sorry, maybe I didn't understand what you meant above.

> 
> > About the wrong cs.archs, I agree, but it seems much easier to rely on the `configs` now that I believe that we could rip it off. What do you think?
> 
> I suspect we still use `cs.archs` somewhere else. If so, isn't ripping it off much of a bigger job than just updating it to a correct value during setup?

I think so, but the code changed a lot recently, and I found out that we have a helper to do exactly what I wanted, so I just ended up using it.

> 
> > The current change fixed my issue and passed the tests, so for now it seems correct (I hope so).
> 
> Sure thing. I understand that it fixes the `#ifdef` emission issue. What I wanted to bring up is that if `cs.archs` is wrong, as you confirmed, then there might be a problem wherever we are still using such field.

For sure there are other places still using `cs.archs`, and I believe they also need some changes. But for my current issue, this fixed it. And for the others, well, we might be hitting them on the CI already, or if not, they will bit us in the future as well.

I truly believe that this is an improvement for now, and we might revisit the other cases of usage of `cs.archs` or other issues, but it's up to you to think about fixing all the cases or merging them in separated PRs as we find other issues. Thanks a lot for the careful review!

klpbuild/klplib/utils.py
vmezzela: but two sets are be equal no matter their order: doing `set(sorted(set()))`, has not effect.

That's why I am proposing to return directly `conf_arch`, that is already a set.

klpbuild/klplib/templ.py
fgyanz: Maybe this would be a good time to clarify the archs design. It seems the concept of `cs.archs` is getting blurrier as time goes on. To give some context, I added it to keep track of each codestream supported archs conditioned by the CONFIGs values during scan/setup + the `--arch` mask. Well, in theory, because that's no longer the case. We have `affected_archs()` doing exactly the first part, which seems to overlap with the original point of having `cs.archs`. Unless I'm missing something, we should set `cs.archs = affected_archs()`, and keep using `cs.archs`. Just  my 2 cents to the discussion :)

marcosps: I think that it solves the issue. Where do you think it's the most suitable place to add this attribution, without adding more problems to the existing checks?

fgyanz: The current flow during `scan\setup` is:

1 - For each codestream, mask `get_default_archs` with `--archs`  and set `cs.archs`.
2 - Print the affected archs for all codestreams. ( The intersection between all codestreams by calling `affected_archs`)
3 - Store the archs intersection in the global `arch` field.

As things stand, neither `cs.archs` nor `affected_archs` really show the whole forest.  
But `cs.archs = cs.archs & affected_archs(cs)` would work, I believe. 

I propose slightly modifying `affected_archs` or having a new function that does the above operation for each codestream. It needs to happen only once on setup or scan, and from then on we can use cs.archs without concern.

marcosps: Done! I hope you guys like the idea :)

fgyanz: Perfect! Hope it worked for you. Thanks!

vmezzela: Yes Marcos, thanks a lot for working on it!

PR #198, opened 2026-03-05
templ: move livepatch header above klp-ccp extracted code

vmezzela:
On sle16 -Werror=missing-prototypes requires the prototype of the klpp_* function. Moving the header before solves the issue.

This is leading to build errors and the CI fails on some LP.


marcosps (approved):
I was myself adjusting the header order in the 16.0 codestreams. Thanks a lot for fixing it!

fgyanz (approved):
LGTM! Thanks. Apparently the CI was canceled (?). Re-triggered. I'll merge once the CI test pass.

PR #199, opened 2026-03-11
Fix some TODO entries

marcosps:
Very simple and direct ones. The first one was an issue that I found while I interrupted the `prepare-tests` and the other ones were easier enough to be done.

### DIFF
--- a/klpbuild/plugins/prepare_tests.py
+++ b/klpbuild/plugins/prepare_tests.py
@@ -114,11 +114,12 @@ def prepare_tests(lp_name, lp_filter):
                 continue
 
             # TODO: there will be only one rpm, format it directly
-            rpm = os.listdir(rpm_dir)
-            if len(rpm) > 1:
-                raise RuntimeError(f"ERROR: {cs.full_cs_name()}/{arch}. {len(rpm)} rpms found. Excepting to find only one")
 
-            for rpm in os.listdir(rpm_dir):
+            rpms = [rpm for rpm in os.listdir(rpm_dir) if rpm.endswith(".rpm")]
+            if len(rpms) > 1:
+                raise RuntimeError(f"ERROR: {cs.full_cs_name()}/{arch}. {len(rpms)} rpms found. Excepting to find only one")
+
+            for rpm in rpms:
                 # Check for dependencies
                 validate_livepatch_module(cs, arch, rpm_dir, rpm)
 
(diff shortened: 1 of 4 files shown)


klpbuild/plugins/prepare_tests.py
```diff
@@ -109,12 +121,7 @@ def prepare_tests(lp_name, lp_filter):
                 logging.info("%s/%s: rpm dir not found. Skipping.", cs.full_cs_name(), arch)
                 continue
 
-            # TODO: there will be only one rpm, format it directly
-            rpm = os.listdir(rpm_dir)
-            if len(rpm) > 1:
-                raise RuntimeError(f"ERROR: {cs.full_cs_name()}/{arch}. {len(rpm)} rpms found. Excepting to find only one")
```
vmezzela, line 119: Is it intended to drop the check here?

vmezzela (commented):
besides the comment, everything looks good!

klpbuild/plugins/prepare_tests.py
marcosps: I think so, yes. It changed quite a bit after it was introduced, but then we will only have one rpm here, so the glob will returns `all`, but in fact only one iteration. It's being done this way so we don't have to parse rpm names, which is even worse to do.

vmezzela: > then we will only have one rpm here

I am actually hitting that `RunTimeError` from time to time, as more than one `rpm` are present in the directory. I don't know if that's a glitch of `IBS` or something caused by `klp-build`. 

But it can happen that more `rpms` are there, and for this reasons I would prefer to keep that warning to avoid preparing a test with the wrong rpm (perhaps coming from a previous build).

vmezzela: I wouldn't parse the rpms name either, as you say. But I would still raise the current warning in case more than one rpm is found.

klpbuild/klplib/ksrc.py
```diff
@@ -239,7 +239,6 @@ def get_patches(cve, savedir=None):
 
     for bc, mbranch in KERNEL_BRANCHES.items():
         # Skip 16.0rt as it will have the same patches of 16.0
-        # TODO: Find a better way to handle this situation
```
fgyanz, line 242: are we sure about this? The current approach is a bit hacky, no? :/

marcosps: Well... I'm not sure how would we handle this differently =/

fgyanz (approved):
LGTM! Thanks!

vmezzela (approved):
lgtm!

PR #201, opened 2026-03-16
data: add arguments --cleanup-obsolete and --update to the command `klp-build data`

ldzhong:
Added a new option --local to the command `klp-build clean-sources` to cleanup the obsolete rpm packages and the corresponding extracted directories, otherwise I have to expand my VM disk image continuously with new kernel MU is released.

### DIFF
--- a/klpbuild/klplib/data.py
+++ b/klpbuild/klplib/data.py
@@ -4,10 +4,69 @@
 # Authors: Vincenzo Mezzela <vincenzo.mezzela@suse.com>
 
 import logging
+import shutil
 
-from klpbuild.klplib.utils import classify_codestreams_str
+from klpbuild.klplib.utils import classify_codestreams_str,ARCHS
 from klpbuild.klplib.ibs import download_cs_rpms
+from klpbuild.klplib.config import get_user_path
+from klpbuild.klplib.kernel_tree import  cleanup_obsolete_trees
 
+def filter_obsolete_directories(target_path, valid_kerns):
+    """
+    check the path including the kernel version is valid or not
+    """
+    ret_paths = []
+    
+    if not target_path.exists():
+        return []
+
+    # Iterate through subdirectories/files in the target path
+    for item in target_path.iterdir():
+        item_str = item.name # Use the name of the folder, not the full path string
+
+        # Check if any valid kernel version string exists within this folder name
+        is_valid = any(k in item_str for k in valid_kerns)
+        if not is_valid:
+            ret_paths.append(item)
+
+    return ret_paths
+
+def cleanup_obsolete_data(valid_codestreams):
+    """
+    Remove obsolete rpm packages and extracted sources
+    """
+    obsolete_paths = []
+
+    valid_kerns = [cs.kernel for cs in valid_codestreams]
+    # The paths we going through:
+    # $data_dir/kernel-rpms
+    # $data_dir/$arch/boot/
+    # $data_dir/$arch/lib/modules
+    # $data_dir/$arch/usr/lib/modules
+
+    data_dir = get_user_path("data_dir")
+    for arch in ARCHS:
+        for d in ["boot", "lib/modules", "usr/lib/modules"]:
+            dest_dir = data_dir/arch/d
+            if dest_dir.exists():
+                obsolete_paths.extend(filter_obsolete_directories(dest_dir, valid_kerns))
+    # check the special dir
+    rpm_dir = data_dir/"kernel-rpms"
+    if rpm_dir.exists():
+        obsolete_paths.extend(filter_obsolete_directories(rpm_dir, valid_kerns))
+
+    # do the cleanup
+    for p in obsolete_paths:
+        try:
+            if p.is_dir():
+                shutil.rmtree(p)
+            else:
+                p.unlink() # Delete file if it's an RPM and not a directory
+            logging.info(f"Removed obsolete path: {p}")
+        except Exception as e:
+            logging.error(f"Error removing {p}: {e}")
+
+    cleanup_obsolete_trees(valid_codestreams)
 
 def download_missing_cs_data(codestreams):
     cs_to_download = __get_cs_missing_data(codestreams)
--- a/klpbuild/klplib/kernel_tree.py
+++ b/klpbuild/klplib/kernel_tree.py
@@ -234,12 +234,15 @@ def __get_active_worktrees(kernel_tree):
 
 
 def __remove_worktree(kernel_tree, worktree_dir):
-    subprocess.check_output(["/usr/bin/git", "-C", kernel_tree, "worktree",
-                             "remove", worktree_dir, "-f", "-f"],
-                            stderr=subprocess.PIPE)
+    try:
+        subprocess.check_output(["/usr/bin/git", "-C", kernel_tree, "worktree",
+                                 "remove", worktree_dir, "-f", "-f"],
+                                stderr=subprocess.PIPE)
 
-    if os.path.isdir(worktree_dir):
-        shutil.rmtree(worktree_dir)
+        if os.path.isdir(worktree_dir):
+            shutil.rmtree(worktree_dir)
+    except Exception as e:
+        logging.info(f"Error removing {worktree_dir}: {e}")
 
 
 def __prune_worktrees(kernel_tree):
--- a/klpbuild/klplib/kernel_tree.py
+++ b/klpbuild/klplib/kernel_tree.py
@@ -161,6 +161,20 @@ def init_cs_kernel_tree(kernel_version, outdir):
             "add", outdir, "--checkout", kernel_tree_git_tag
         ], stderr=subprocess.PIPE)
 
+def cleanup_obsolete_trees(codestreams):
+    """
+    Remove obsolete kernel source trees.
+    """
+    kernel_tree = get_user_path("kernel_dir")
+    worktrees = __get_active_worktrees(kernel_tree)
+    valid_kerns = [cs.kernel for cs in codestreams]
+
+    for wt in worktrees:
+        wt_str = Path(wt).name # Use the name of the folder, not the full path string
+        is_valid = any(k in wt_str for k in valid_kerns)
+        if not is_valid:
+            logging.info("Removing worktree %s", wt)
+            __remove_worktree(kernel_tree, wt)
 
 def cleanup_kernel_trees():
     """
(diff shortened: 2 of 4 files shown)


klpbuild/klplib/kernel_tree.py
```diff
@@ -185,6 +185,65 @@ def cleanup_kernel_trees():
...
+
+    return ret_paths
+
+def cleanup_local_sources():
+    """
+    Remove obsolete rpm packages and extracted sources
+    """
+    obsolete_paths = []
+
+    from klpbuild.klplib.supported import get_supported_codestreams
```
vmezzela: maybe this can be moved with the other imported modules at the beginning of the file?

klpbuild/klplib/kernel_tree.py
```diff
@@ -185,6 +185,65 @@ def cleanup_kernel_trees():
...
+    all_codestreams = get_supported_codestreams()
+    valid_kerns = [cs.kernel for cs in all_codestreams]
+    # The paths we going through:
+    # $data_dir/kernel-rpms
+    # $data_dir/$arch/boot/
+    # $data_dir/$arch/lib/modules
+    # $data_dir/$arch/usr/lib/modules
+    # $data_dir/$arch/usr/src only present for x86_64
```
vmezzela: Cleaning up the expanded kernel trees under `$data_dir/x86_64/usr/src` would leave the worktrees in the `$kernel_tree` git directory dangling.

vmezzela (commented):
Hi Lidong, thanks a lot for the PR! It's a useful feature to have!
I only have a remark: `klplib/kernel_tree.py` and `plugins/cleanup-sources.py` are meant to handle only the expanded kernel trees under `$data_dir/x86_64/usr/src`, while your PR cleanup everything is obsolete (including kernel objects/configs), which is even better.

To remain consistent with the current structure of the code, I would suggest to move part of the code you implemented to `plugins/data.py` and run it with a `--cleanup-obsolete` argument.

klpbuild/klplib/kernel_tree.py
vmezzela: If we move the code under `plugins/data.py`, then we can implement a function that is only meant to cleanup the kernel trees in `klplib/kernel_tree.py` and removes them using `git worktree remove` , similar to what `cleanup_kernel_trees()` does but only obsolete tree.

Or you can also modify directly `cleanup_kernel_trees()` to take an optional `valid_kerns` and remove only obsolete ones.

fgyanz:
Hi Lidong, thanks a lot for the PR! This feature has been in our TODO backlog for quite some time now.  I second @vmezzela comments except for one detail:

> To remain consistent with the current structure of the code, I would suggest to move part of the code you implemented to plugins/data.py and run it with a --cleanup-obsolete argument.

Yes, but perhaps it would be simpler to have the cleanup enabled by default on `data --download`?  There are very few cases in which it makes sense to not remove obsolete data. If we go with this route, I'd propose renaming the command to `data --update`, like any package manager. What do you think?

vmezzela:
> If we go with this route, I'd propose renaming the command to `data --update`, like any package manager. What do you think?

perhaps we keep `--download` to only download the data, `--cleanup-obsolete` to only cleanup obsolete kernel objects/trees, and `--update` to do both.

I personally prefer to have more fine grained control. What do you think?

fgyanz:
> I personally prefer to have more fine grained control. What do you think?

Sure, ideally that's the way to go. However, just to do the devil's advocate: the more arguments we add, the complex the codebase and documentation. If there are valid use cases, then I guess it's justified.

vmezzela:
Cleaning up obsolete data without downloading anything new is one use case that I would find useful

ldzhong:
Thanks for your reviews. I have made some changes to the code based on your suggestions and adjusted the subject of this PR.

klpbuild/klplib/data.py
```diff
@@ -4,10 +4,71 @@
...
+
+    valid_kerns = [cs.kernel for cs in codestreams]
+    # The paths we going through:
+    # $data_dir/kernel-rpms
+    # $data_dir/$arch/boot/
+    # $data_dir/$arch/lib/modules
+    # $data_dir/$arch/usr/lib/modules
+
+    data_dir = get_user_path("data_dir")
+    archs = ["s390x","ppc64le","x86_64"]
```
marcosps: We have a global variable called ARCHS where we have all the current available architectures so it can cover aarch64 in the future.

klpbuild/klplib/data.py
```diff
@@ -4,10 +4,71 @@
...
+    dirs = ["boot", "lib/modules", "usr/lib/modules"]
```
marcosps: dirs and archs are defined and used once. I believe that you can use them directorly in the loop declarations.

marcosps (approved):
Thats a very good feature, thanks a lot for working on it! I have only minimal comments about the code style, besides that the change is good.

ldzhong:
> Thats a very good feature, thanks a lot for working on it! I have only minimal comments about the code style, besides that the change is good.

Thanks for your reply, Marcos. I updated the code based on your review.

klpbuild/plugins/data.py
```diff
@@ -33,5 +36,13 @@ def run(download, force, lp_filter):
...
+        cleanup_obsolete_sources(supported_codestreams)
+    elif update:
+        if force:
+            download_cs_data(filtered_codestreams)
+        else:
+            download_missing_cs_data(filtered_codestreams)
+        cleanup_obsolete_sources(supported_codestreams)
     else:
-        logging.error("Use --download")
+        logging.error("Use one of --download, --cleanup-obsolete and --update for this command")
```
vmezzela: s/and/or/

klpbuild/klplib/data.py
```diff
@@ -4,10 +4,69 @@
...
+        item_str = item.name # Use the name of the folder, not the full path string
+
+        # Check if any valid kernel version string exists within this folder name
+        is_valid = any(k in item_str for k in valid_kerns)
+        if not is_valid:
+            ret_paths.append(item)
+
+    return ret_paths
+
+def cleanup_obsolete_sources(codestreams):
```
vmezzela: perhaps it would be more correct to call this function `cleanup_obsolete_data`, since it doesn't remove the sources only.

Also, to make it even more clear, I would call the argument `valid_codestreams`  rather than the generic `codestreams`. The same comment applies to some other function where this change can be applied.

klpbuild/klplib/data.py
```diff
@@ -4,10 +4,69 @@
...
 import logging
+import shutil
 
-from klpbuild.klplib.utils import classify_codestreams_str
+from klpbuild.klplib.utils import classify_codestreams_str,ARCHS
 from klpbuild.klplib.ibs import download_cs_rpms
+from klpbuild.klplib.config import get_user_path
+from klpbuild.klplib.kernel_tree import  cleanup_obsolete_trees
 
+def cleanup_dir(target_path, valid_kerns):
```
vmezzela: what about calling this function `filter_obsolete_directories` or something similar?

klpbuild/klplib/data.py
```diff
@@ -4,10 +4,69 @@
...
+    # do the cleanup
+    for p in obsolete_paths:
+        try:
+            if p.is_dir():
+                shutil.rmtree(p)
+            else:
+                p.unlink() # Delete file if it's an RPM and not a directory
+            logging.info(f"Removed obsolete path: {p}")
+        except Exception as e:
+            logging.info(f"Error removing {p}: {e}")
```
vmezzela: Just a nit, but perhaps this should be `logging.error`?

klpbuild/plugins/data.py
```diff
@@ -33,5 +36,13 @@ def run(download, force, lp_filter):
...
         else:
             download_missing_cs_data(filtered_codestreams)
+    elif cleanup_obsolete:
+        cleanup_obsolete_sources(supported_codestreams)
+    elif update:
+        if force:
+            download_cs_data(filtered_codestreams)
+        else:
+            download_missing_cs_data(filtered_codestreams)
+        cleanup_obsolete_sources(supported_codestreams)
```
vmezzela: Mutually exclusive group not enforced here: `--download`, `--cleanup-obsolete`, and `--update` are handled as elif branches, but nothing prevents the user from passing `--download --cleanup-obsolete` together. With elif, `--download` would silently win and `--cleanup-obsolete` would be ignored. I think argparse provides some way to handle mutually exclusive option, however I don't mind if we leave this as is. I think if this ever happens it's not an issue.

vmezzela (changes requested):

klpbuild/plugins/data.py
```diff
@@ -18,13 +18,16 @@ def register_argparser(subparser):
...
     add_arg_lp_filter(fmt)
-    fmt.add_argument("--download", required=True, action="store_true",
+    fmt.add_argument("--download", required=False, action="store_true",
                      help="Download all the missing supported codestreams data")
     fmt.add_argument("--force",action="store_true",
                      help="Re-download also codestream that are not missing")
+    fmt.add_argument("--cleanup-obsolete", required=False, action="store_true",
+                     help="Cleanup the obsolete files and directories")
+    fmt.add_argument("--update", required=False, action="store_true",
+                     help="Do the job of --download and --cleanup-obsolete")
```
fgyanz, line 28: Do you mind also adding the new options to the man page? The file would be: https://github.com/SUSE/klp-build/blob/main/klp-build.1

Everything else looks good to me. Thanks!

klpbuild/plugins/data.py
ldzhong: how about we remove the option --update? Something like 

>  34     if download:
>  35         if force:
>  36             download_cs_data(filtered_codestreams)
>  37         else:
>  38             download_missing_cs_data(filtered_codestreams)
>  39     if cleanup_obsolete:
>  40         cleanup_obsolete_data(supported_codestreams)
>  41     if not download and not cleanup_obsolete:
>  42         logging.error("Use --download or --cleanup-obsolete for this command")

vmezzela:
```py
    if download or update:
        if force:
            download_cs_data(filtered_codestreams)
        else:
            download_missing_cs_data(filtered_codestreams)
    if cleanup_obsolete or update:
        cleanup_obsolete_data(supported_codestreams)
    if not (download or cleanup_obsolete or update):
        logging.error("Use --download, --cleanup-obsolete or --update for this command")

```

What about this? so we can keep the `--update` as well.

ldzhong: Sure. I have made the change as you suggested here in my latest PR.

ldzhong:
I have updated my PR based on your feedback. Thanks for your suggestions.

vmezzela:
Thanks a lot @ldzhong ! Merged!

PR #203, opened 2026-03-17
CVE arg

marcosps:
This PR makes the --cve to go away. The only disadvantgage right now is that we can't really use random names for a bsc, since we rely on the bsc to get the CVE. The only change here is that, if we use --no-check, then the bsc doesn't matter, because we don't run scan, and so we don't need to CVE number.


klpbuild/plugins/setup.py
```diff
@@ -161,7 +155,8 @@ def setup_codestreams(lp_name, data):
         all_codestreams = get_supported_codestreams()
         codestreams = utils.filter_codestreams(data["lp_filter"], all_codestreams)
     else:
-        _, upstream, patched_cs, codestreams = scan(data["cve"], data["conf"],
+        _cve = bugzilla.get_bug_cve(bugzilla.get_bug(lp_name.replace("bsc", "")))
```
fgyanz: I'd propose having `lp_name.replace("bsc", "")` inside bugzilla:get_bug. We mostly use "bsc..." string within klp-build, but bugzilla requires just the id values. So stripping off the "bsc" automatically makes sense imho.

Also, why use `_cve` as variable name instead of plain `cve`?

fgyanz (commented):
Just some small nitpicking from my side. Thanks a lot for adding this!

klpbuild/plugins/setup.py
marcosps: Thanks for the suggestion of removing the bsc prefix on bugzilla. Done here locally!

The `_cve` variable is because later on there is a call to set_codestreams_data, which already has an argument `cve`. But either way I changed it back to `cve`. Thanks!

fgyanz (approved):
LGTM!

PR #204, opened 2026-03-19
templ: fix regexp to find incomplete struct declaration

vmezzela:
The regexp is broken when there are more than one '*':

Given the input 'struct nlattr **tca', the regexp:

- before this patch matched 'struct nlatt'
- now matches 'struct nlattr'


fgyanz (approved):
Makes sense! LGTM!

marcosps (approved):
Would it be nice if we have a test for it? Maybe extracting this functionatlity into a different function and call it from a test.

Either way, that fixes an issue, so we merge it and think about a test in the future. Thanks a log!

PR #206, opened 2026-03-19
extract: keep original output during extract

vmezzela:
Storing the initial output of klp-ccp allows to detect nasty regexp bug on the cleanup side by checking the diffs.


marcosps (approved):
This is something necessary whenever we are in doubt about if we messed up with something or not. Thanks you very much for this PR!

fgyanz (approved):
Agreed! Very useful :) LGTM

PR #208, opened 2026-03-23
setup: assert and log CVE retrieved from bugzilla

vmezzela:
Just a minor adjustment after merging #203 as I find the log useful.


fgyanz (approved):
Makes sense. LGTM!

marcosps (approved):
LGTM!

PR #209, opened 2026-03-23
setup: add --add-patches option to specify prerequisite patches

vmezzela:
It happens often that the extract command fails to apply the backports on the kernel sources. This usually happens because there are missing patches that needs to be pulled in and applied to the source before the backports to prepare the code context. This PR introduces the `--prereq` option to instruct klp-build to grab more patches and apply them before the backports.

### DIFF
--- a/klp-build.1
+++ b/klp-build.1
@@ -189,6 +189,13 @@ will be livepatched instead.
 .TP
 .BI --archs " {ppc64le,s390x,x86_64} [{ppc64le,s390x,x86_64} ...]"
 Supported architectures for this livepatch.
+.TP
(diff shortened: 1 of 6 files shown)


vmezzela:
Force pushed to change `--prereq` to `--add-patches`.

marcosps (approved):
LGTM! I believe that @fgyanz will also ask for the new option to be included in the man page :)

vmezzela:
> LGTM! I believe that @fgyanz will also ask for the new option to be included in the man page :)

Oh right, let me add it :P

fgyanz (approved):
LGTM! Thanks!

PR #211, opened 2026-03-24
format-patches: don't bail out when ccp groups are missing

vmezzela:
If for any reason I dropped the `ccp` directory, then I cannot run `format-patches` as it requires the `groups` file.


fgyanz:
Thanks! The changes are sound and good enough as is. However, I would like to propose a different approach in case you hadn't thought about it: As a fallback, derive the groups from the livepatch branches name. This is an uncommon case, so I don't know if it's worth the effort. So, again, I'm perfectly okay merging as is :)

vmezzela:
Right, that's also another option!

fgyanz (approved):

marcosps (approved):
It's  a valid use case. LGTM!

PR #214, opened 2026-03-30
scripts/run-kgr-test: Remove /home/nstange fallback path

fgyanz:


vmezzela (approved):
lgtm!

marcosps (approved):
I believe that we can start trimming down the run-kgr-test script, since we are defaulting to store things on /home/klp, but this can be done in the future!

LGTM!

PR #215, opened 2026-04-08
Check livepatch EOL

fgyanz:
Print the livepatch EOL (End Of Life) during `scan` and warn if all the affected codestreams will become unsupported in less than 30 days. In such case, the developer might choose to skip working on this CVE.


vmezzela (approved):
lgtm!

marcosps (approved):
LGTM, thanks!

PR #216, opened 2026-04-09
Fix pylint reported warnings and errors

fgyanz:
The github CI tests were not executed on recent PRs because the pylint check was failing before that point. There were many new warning that had lowered the quality score to 9.77/10 , while the threshold limit is set at 9.83. This PR fixes as much warnings and errors as possible in order to reach the threshold again.

Fixes:
- Use UPPER_CASE for module-level mutable state variables
- Replace dangerous mutable default arguments with None
- Fix logging f-string interpolation to use lazy % formatting
- Narrow broad Exception catches to specific types
- Fix bad indentation, trailing whitespace/newlines
- Rename loop variable shadowing 'patch' module import
- Disable global-statement (legitimate lazy-init pattern)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -94,6 +94,7 @@ disable=raw-checker-failed,
         wrong-import-position,
         unspecified-encoding,
         fixme,
+        global-statement,
 
 # Enable the message, report, category or checker with the given id(s). You can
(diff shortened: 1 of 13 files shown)


marcosps (approved):
Thanks a lot for improving it!

vmezzela (approved):

PR #218, opened 2026-04-10
klplib/kernel_tree: git-fs integration

fgyanz:
The current extraction phase creates one git-worktree per affected kernel/codestream to parallelize klp-ccp runs. Each worktree consumes 1-2 GB on disk, and with up to 120 supported codestreams, total disk usage can reach 240 GB.

The main goal of this PR is to replace worktrees with [`git-fs`](https://github.com/fgyanz/git-fs); a read-only FUSE-based git filesystem mounted over the kernel repository. Source files for each kernel version are served directly from the git object store, eliminating full on-disk checkouts and drastically reducing storage overhead.

The PR cannot be merged until `git-fs` development is fully finished and distributed as a package in OpenSUSE.


fgyanz:
@vmezzela @marcosps I think this one is stable enough. Feel free to review it :)

marcosps:
The first commit uses `get_repos_dir`, which is added in the second patch. other that it LGTM

fgyanz:
I've done a small change so that `git-fs` integration works seamlessly with PR #266

vmezzela (approved):
Hi @fgyanz , I have done a first review of this code. It's awesome! I will test it further while reviewing some livepatches before merging it! thanks a lot!

klpbuild/klplib/kernel.py
```diff
@@ -0,0 +1,382 @@
...
+            __KERNEL_FS = fs_path
+            return
+
+        gitfs_bin = shutil.which("git-fs")
+        if not gitfs_bin:
+            logging.warning(
+                "git-fs binary not found. Install git-fs to enable "
+                "FUSE-based kernel tree access."
+            )
+            return
```
vmezzela: Should we bail here?

klpbuild/klplib/kernel.py
```diff
@@ -0,0 +1,382 @@
...
+def __fini_kernel_fs():
+    """
+    Unmount the git filesystem on exit().
+    """
+
+    if not __KERNEL_FS or not os.path.ismount(__KERNEL_FS):
+        return
+
+    subprocess.check_output(
+        ["git", "fs", "-u", str(__KERNEL_FS)],
```
vmezzela, line 145: is it better to use `shutil.which("git-fs")` as we do during the mount?

klpbuild/klplib/kernel.py
```diff
@@ -0,0 +1,382 @@
...
+        kernel_version (str): Kernel version to check.
+        file_path (str): Path of the file to verify.
+
+    Returns:
+        True if exists. False otherwise.
+    """
+    return (get_kernel_tag_path(kernel_version) / file_path).exists()
+
+
+def read_file_in_tag(kernel_version, file_path):
```
vmezzela, line 284: Missing `@__check_kernel_tags_are_fetched` decorator?

klpbuild/klplib/codestream.py
```diff
@@ -144,24 +144,24 @@ def __eq__(self, cs):
     def __str__(self):
         return self.full_cs_name()
 
-    def get_src_dir(self, arch=None, init=True):
+    def get_src_dir(self):
+        return get_kernel_tag_path(self.kernel)
+
+    def get_build_dir(self, arch=None):
```
vmezzela: What is the build dir? it seems to me that this returns the path to the old source directory, which should no longer exist with this PR.

klpbuild/klplib/kernel.py
```diff
@@ -0,0 +1,382 @@
...
+    global __KERNEL_FS
+
+    with __kernel_fs_init_lock:
+        if __KERNEL_FS:
+            return
+
+        fs_path = get_repos_dir() / TREE_NAME / "fs"
+
+        if os.path.ismount(fs_path):
+            logging.warning("%s: already mounted", str(fs_path))
```
vmezzela, line 108: is that okay if such a thing happen? isn't it kind of a red flag? e.g. a git-fs in use by someone else

vmezzela (commented):
Hi @fgyanz, I left some comments inline and I have two questions:
1. Is git-fs already packages? (though I am okay to merge even if not)
2. Does the CI needs to be updated?

klpbuild/klplib/codestream.py
fgyanz: build_dir points to the src/ dir, yes. And we no longer have the kernels sources there, true. But the src/ still exists because it contains the ipa-clones, the `compile-commands.json` etc within each `linux-*-obj`. So, in that sense, the directory no longer contains any source code.   In that directory is where we also run `make --dry-run` . Due to the previous reasons, I felt that the `get_src_dir` name would be misleading, so I proposed the rename to get_build_dir (any better ideas?).

Unfortunately, the directory itself cannot be renamed easily because several rpms expect that path during extraction, as we discussed yesterday.

klpbuild/klplib/kernel.py
fgyanz: Is it a red flag? Yes and no. The filesystem could already exist as a leftover from a previous klp-build process that crashed badly. In such case it would be safe to use.
It could also be from another process still running, yes. But it would be still safe to use. The worst that can happen is that the filesystem gets unmounted by someone else. In such case, the next read operation will gracefully fail. But I'd rather warn first and let the user know rather than error from the start. What do you think?

klpbuild/klplib/kernel.py
fgyanz: So, my logic is: if `git-fs`  binary exists (checked during mount), then we can assume that it can be called through `git`, as any other `git` command. Do you think it would be better to use shutil ?

fgyanz:
Hi @vmezzela , thanks for reviewing it.

> Is git-fs already packages? (though I am okay to merge even if not)

Not yet. I was waiting until the PR passed the review phase. It shouldn't take much time to package it (famous last words!).

> Does the CI needs to be updated?

In which sense? The CI has been using this PR since April. Is there anyhting to update?

klpbuild/klplib/kernel.py
vmezzela: I'd just prefer to use the same pattern everywhere. The current code does both `shutil` and `git fs`. Unless there's a reason I do not see. But yeah this detail is a non blocker of course

vmezzela:
> > Does the CI needs to be updated?
> 
> In which sense? The CI has been using this PR since April. Is there anyhting to update?


sorry, I meant the github CI :P

fgyanz:
> > > Does the CI needs to be updated?
> > 
> > 
> > In which sense? The CI has been using this PR since April. Is there anyhting to update?
> 
> sorry, I meant the github CI :P

Ah!! My bad, I always default to our CI :D 

Right, our github CI needs to have `git-fs` installed. My idea was to create the obs package
and install it through zypper rather than installing from sources.

klpbuild/klplib/codestream.py
vmezzela: > build_dir points to the src/ dir, yes. 

Sorry, maybe I was not clear. The current `get_build_dir()` returns the path to the old source directory. With this I meant `$data/x86_64/usr/src/linux-<kver>` and not the generic `$data/x86_64/usr/src`.

> And we no longer have the kernels sources there, true. 

Exact!

> But the src/ still exists because it contains the ipa-clones, the `compile-commands.json` etc within each `linux-*-obj`. So, in that sense, the directory no longer contains any source code. 

Exact!

Let me explain better: I would drop `get_build_dir()` because it returns the `$data/x86_64/usr/src/linux-<kver>` (which doesn't exists), and currently this is only used by `get_obj_dir()` by appending the `-obj` suffix.

> In that directory is where we also run `make --dry-run` . Due to the previous reasons, I felt that the `get_src_dir` name would be misleading, so I proposed the rename to get_build_dir (any better ideas?).

`make --dry-run` is not run in `get_build_dir()` but in `get_obj_dir()` and as said before, `get_obj_dir()` uses `get_build_dir()` just to build the path.

See:

```py
   def get_build_dir(self, arch=None):
        # Before sle16, only -rt codestreams have a suffix for source directory
        has_rt_suffix = self.rt and self.sle < 16
        name = self.get_full_kernel_name() if has_rt_suffix else self.kernel

        if not arch:
            arch = preferred_arch([self])

        return get_datadir(arch)/"usr"/"src"/f"linux-{name}"

    def get_obj_dir(self, arch=None):
        if not arch:
            arch = preferred_arch([self])

                             vvvvvvvvvvv
        return Path(f"{self.get_build_dir(arch)}-obj", arch, self.get_kernel_type())
```

And given that `get_obj_dir()` is the only caller, we could just move the body of `get_build_dir()` into it. 

What do you think?

klpbuild/klplib/kernel.py
vmezzela: You know it better, I trust you! Ack from my side.

vmezzela:
> > > > Does the CI needs to be updated?
> > > 
> > > 
> > > In which sense? The CI has been using this PR since April. Is there anyhting to update?
> > 
> > 
> > sorry, I meant the github CI :P
> 
> Ah!! My bad, I always default to our CI :D
> 
> Right, our github CI needs to have `git-fs` installed. My idea was to create the obs package and install it through zypper rather than installing from sources.

I see.

Do you prefer to package it before the merge? IMHO we can merge before but we should guarantee that the package and the github CI are ready before merging to the `main` branch.

klpbuild/klplib/codestream.py
fgyanz: ah! Now I understand. Makes sense. I can move get_build_dir into get_obj_dir! :)

vmezzela: Thanks a lot!

fgyanz:
> Do you prefer to package it before the merge? IMHO we can merge before but we should guarantee that the package and the github CI are ready before merging to the main branch

Yeah, I would rather create the package and update the CI now than before a release. Less risky imho.

vmezzela:
> > Do you prefer to package it before the merge? IMHO we can merge before but we should guarantee that the package and the github CI are ready before merging to the main branch
> 
> Yeah, I would rather create the package and update the CI now than before a release. Less risky imho.

Sure, but not clear if you want to create the package before or after the merge into `devel`

fgyanz:
Before merging into `devel` :)

klpbuild/klplib/codestream.py
fgyanz: done!

klpbuild/klplib/kernel.py
fgyanz: Fair enough. I update the `mount` operation to use `git`. The `shutil` now will only be used to verify that gitfs exists. But for nothing else.

fgyanz:
@vmezzela all CI checks have passed. From now on, `git-fs` can be downloaded from: `https://download.opensuse.org/repositories/home:/fgonzalez/openSUSE_Tumbleweed/home:fgonzalez.repo`

PR #219, opened 2026-04-13
extract: Handle new IPA-removed symbol warning 

fgyanz:
klp-ccp now warns when a symbol is found in the ELF but not in the IPA. Catch the warning and print a more sane
log for the klp-build user.

Slightly refactor the extract code to avoid redundant parsing of klp-ccp.out.txt. Add tests covering the new changes.

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -694,6 +694,34 @@ def cmd_args(lp_name, cs, fname, out_dir, fdata, cmd, avoid_ext):
     return ccp_args, env
 
 
+def _warn_optimized_clone(match, cs, fname):
+    opt_symbol_name = match.group(1)
+    symbol_name = opt_symbol_name.split(".")[0]
+    logging.warning("%s:%s: Symbol %s contains optimized clone: %s",
+                    cs.full_cs_name(), fname, symbol_name, opt_symbol_name)
+    logging.warning("Make sure to patch all the callers of %s.", symbol_name)
+
+
+def _collect_dup_symbol(match, cs, fname):
+    cs.files[fname].setdefault("dup_symbols", [])
+    cs.files[fname]["dup_symbols"].append(match.group(1))
+
+
+CCP_WARNING_SPECS = [
+    (r'warning: optimized function "([^"]+)" in callgraph', _warn_optimized_clone),
+    (r'conflicting definitions for symbol "([\w_]+)" found in ELF', _collect_dup_symbol),
+]
+
+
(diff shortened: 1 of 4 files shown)


klpbuild/plugins/extract.py
```diff
@@ -702,13 +702,20 @@ def _warn_optimized_clone(match, cs, fname):
     logging.warning("Make sure to patch all the callers of %s.", symbol_name)
 
 
+def _warn_ipa_removed(match, cs, fname):
+    symbol_name = match.group(1)
+    logging.warning("%s:%s: Unable to verify if %s symbol is inlined or not. Skipping it.",
```
vmezzela: What should a user do when this warning is raised? 

I think that from the log message is not clear what happened: should we either retain a message similar to the klp-ccp log to explain it? Or maybe explain directly what the user should do in this case. What do you think?

fgyanz: Hi @vmezzela ,

> I think that from the log message is not clear what happened: should we either retain a message similar to the klp-ccp log to explain it? Or maybe explain directly what the user should do in this case. What do you think?

I personally think that the klp-ccp log is too technical to be used directly. It's fine for someone debugging klp-ccp, but for a klp-build user there's no need to really know anything about IPA imho.

> What should a user do when this warning is raised?

Manually verify whether the symbol is inlined or not. If it is, then all good (the `klpp_` symbol is not externalized by default in this case). If it is not inlined, they will have to manually add it to the livepatch header file and `patched_funcs.csv`, handling correctly conflicting symbols.

I didn't include any additional instructions because the average klp-build user already knows the implication of an `inlined` / `not inlined` symbol in the context of livepatching, and how to deal with it. Well, at least that's my reasoning. But if you think it's not clear enough, feel free to propose alternative sentences!

vmezzela: Your reasoning make sense.

One last doubt is: what does it mean "skipping it"? Is any verification being skipped in klp-ccp?

fgyanz: No, just that klp-build will treat this symbol as inlined. Hence "skipped", as "symbol won't be livepatched". Maybe I should reword it to something more clear... It's just that I don't want to be too verbose.

vmezzela: Mmmh, okay. Maybe something that prompts the user to double check manually? "Skipping it" make me think that the warning is already addressed by skipping the symbol

fgyanz: done!

vmezzela (approved):
lgtm!

marcosps (approved):
LGTM!

PR #220, opened 2026-04-14
Forwarding enum declarations in livepatch header

fgyanz:
As we already do for structs, this PR adds support for enums:

```
struct macvlan_port;
struct net_device;
struct sk_buff;

# New enum declaration
enum macvlan_mode;

void klpp_macvlan_broadcast(struct sk_buff *skb, const struct macvlan_port *port, struct net_device *src, enum macvlan_mode mode);
```

Note that forwarding enum declarations, as opposed with structs, is not supported by standard C99. Luckily, GCC and Clang do support this extension.

### DIFF
--- a/klpbuild/klplib/templ.py
+++ b/klpbuild/klplib/templ.py
@@ -462,27 +462,32 @@ def __generate_klpp_header(cs):
     '''
     Generate any klpp-specific header information, such as:
     - klpp functions declarations.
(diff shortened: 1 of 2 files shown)


vmezzela (approved):
lgtm!

marcosps (approved):
lgtm!

PR #221, opened 2026-04-15
templ: relax __is_check_enabled to improve codestreams grouping

vmezzela:
If there's a CVE that affects only x86_64, then the #ifdef will be emitted on -default codestreams (cs.archs != cs.get_default_archs()) but not on -rt (in this case cs.archs == cs.get_default_archs()). Therefore the different #ifdef emission decision across codestreams will split the grouping even more.


marcosps (approved):
At first I thought that it would introduce a bug that I fixed in the past, but this fix along with a891969c07a1e95371ad4dda586296b1776e41a7 might work, thanks a lot!

fgyanz (approved):
lgtm!

PR #222, opened 2026-04-15
tox: pass KLP_CCP_PYTHONPOLICY_PATH into tox env

vmezzela:
I needed it when using a klp-ccp version not installed in the PATH, but might be worth merging this anyway in case can be useful to you too


marcosps (approved):
lgtm!

fgyanz (approved):
LGTM!

PR #223, opened 2026-04-17
Do not extract vmlinux and kernel modules

marcosps:
This PR solves #202 in a different way: instead of always extracting a module when necessary, we never extract them. Instead, we do it in runtime, so we don`t waste disk space with it.

When setup is running and downloading additional rpms files, it's not uncommon to developers to stop the process in the middle of it, skipping the extracting of kernel modules and vmlinux. The result is a half baked data directory, which requires further klp-build commands in order to have them extracted. This PR resolves this issue, decreasing the possiblity of such issues to happen.

This PR should be merged after only after https://github.com/SUSE/klp-ccp/pull/28 is merged. klp-ccp needs to be able to also decompress the files in runtime.

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -145,19 +145,46 @@ class Codestream:
             result = subprocess.run(cmd, capture_output=True, text=True, check=True)
             return result.stdout
 
-
-    def get_boot_file(self, file, arch=None):
-        assert file.startswith("vmlinux") or file.startswith("config") or file.startswith("symvers")
-
+    def get_boot_dir(self, arch=None):
+        """
+        Return the directory containing the boot files for the current
+        codestream.
+        """
         if not arch:
             arch = preferred_arch([self])
 
         if self.is_slfo:
-            return Path(self.get_mod_path(arch), file)
+            return self.get_mod_path(arch)
 
-        # Strip the suffix from the filename so we can add the kernel version in the middle
-        fname = f"{Path(file).stem}-{self.get_full_kernel_name()}{Path(file).suffix}"
-        return get_datadir(arch)/"boot"/fname
+        return get_datadir(arch) / "boot"
+
+    def get_boot_filename(self, file):
+        """
+        Files like vmlinux, config and symvers can have different names
+        depending which codestream there are. Return the expected names of the
+        files in the current codestream.
+
+        On SLFO, the filename are don't contain any suffixes, like kernel
+        versions, because they live inside usr/lib/modules directory. For
+        older codestreams we need to append the current kernel version to the
+        file, because all config, symvers and vmlinux live in the same /boot
+        directory.
+        """
+
+        if self.is_slfo:
+            return file
+
+        return f"{file}-{self.get_full_kernel_name()}"
+
+    def get_boot_file(self, file, arch=None):
+        assert file.startswith("vmlinux") or file.startswith("symvers")
+
+        mod_path = self.get_boot_dir(arch)
+
+        if not self.is_slfo:
+            file = f"{file}-{self.get_full_kernel_name()}"
+
+        return mod_path / file
 
     def get_repo(self):
         if self.update == 0 or self.is_slfo:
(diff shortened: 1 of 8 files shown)


klpbuild/klplib/codestream.py
```diff
@@ -145,19 +145,31 @@ def get_config_content(self, arch=ARCH):
...
 
         if not arch:
             arch = preferred_arch([self])
 
         if self.is_slfo:
-            return Path(self.get_mod_path(arch), file)
+            mod_path = self.get_mod_path(arch)
+
+        else:
+            # Strip the suffix from the filename so we can add the kernel version in the middle
```
vmezzela: does this comment still apply? It seems a leftover from the previous code.

klpbuild/klplib/utils.py
```diff
@@ -506,3 +506,23 @@ def is_lp_eol_soon(cs_list):
 def validate_lp_name(lp_name):
     if not lp_name.startswith("bsc"):
         raise ValueError("Please use prefix 'bsc' for livepatch name")
+
+
+def find_kernel_module(path_dir, file):
```
vmezzela: Since we use this function to find also vmlinux, config and symvers and not only kernel modules, I think that the current name is misleading. Can we rename it?

klpbuild/klplib/codestream.py
```diff
@@ -415,17 +415,16 @@ def find_obj_path(self, arch, mod):
...
+            obj_match = re.search(rf"([\w\/\-]+\/{mod})\.ko", f.read())
             if not obj_match:
                 raise RuntimeError(f"{self.full_cs_name()}-{arch} ({self.kernel}): Module not found: {mod}")
 
-        # modules.order will show the module with suffix .o, so make sure the extension.
-        obj_path = mod_path/(PurePath(obj_match.group(1)).with_suffix(".ko"))
-        # Make sure that the .ko file exists
-        assert obj_path.exists(), f"Module {str(obj_path)} doesn't exists. Aborting"
+        fpath = Path(obj_match.group(1))
+        fmod = find_kernel_module(Path(mod_path, fpath.parent), fpath.name)
```
vmezzela: From the old code, I understand that is was possible to find module having the `.o` suffix. Is it still the case? because if so, we're no longer covering it as far as I can see.

Also, the whole point of this function was to take advantage of `modules.order` to find the path of a given module, but since you introduced the `find_kernel_module` utility that uses shell globbing to find it, would it make more sense to just replace this function with the logic from `find_kernel_module`? I think they are overlapping.

vmezzela (changes requested):
Hi Marcos, thanks for the PR. Overall it looks good, I would just propose to change one thing: instead of having `find_kernel_module()` in utils.py, we can just drop it and implement it under `find_obj_path()` in the codestream class?

Having both of them I think is not needed, I explained better in the comment I left inline.

klpbuild/klplib/codestream.py
```diff
@@ -390,7 +402,7 @@ def find_obj_path(self, arch, mod):
         # Return the path if the modules was previously found for ARCH, or refetch if
         # the obejct is for a different architecture
         obj = self.modules.get(mod, "")
-        if obj:
+        if obj and mod != "vmlinux":
```
fgyanz: Wouldn't it be simpler to just have
```
         # We already know the path to vmlinux, so return it
         if not is_mod(mod):
             return self.get_boot_file("vmlinux", arch).relative_to(get_datadir(arch))
```
before `self.modules.get` ?

klpbuild/klplib/utils.py
```diff
@@ -493,3 +506,23 @@ def is_lp_eol_soon(cs_list):
...
```
fgyanz: Please create some unit tests covering the new function. It should be doable since it just requires a dir and file name.

klpbuild/klplib/utils.py
```diff
@@ -174,7 +176,18 @@ def get_elf_object(obj):
     with open(obj, "rb") as f:
         data = f.read()
 
-    return ELFFile(io.BytesIO(data))
+    # FIXME: use magic lib instead of checking the file extension
```
fgyanz: Although this came with the reverted commit, why not fix it now? Is there any blocker? AFAIK it should pretty easy to implement.

fgyanz (commented):
Thanks @marcosps ! Just nitpicking from my side.

klpbuild/klplib/codestream.py
marcosps: I need to recheck, but I believe that the current setup/scan returns the path to the module, so maybe we don't even need to read `modules.order` anymore. Let me think about it!

klpbuild/klplib/utils.py
marcosps: Sure, I'll think in a better name for it

marcosps:
@fgyanz @vmezzela now I changed how things work: first I start to remove the usage of get_boot_file, and slowly I`m moving to the point of not decompressing the code anymore. Now `find_kernel_module` was merged into `find_obj_path`, and tests were added to some new and existent functions. Please review!

fgyanz:
@marcosps thanks for working again on this PR! Before reviewing, do you mind rebasing again on devel branch? Looks like the base is not up to date.

marcosps:
> @marcosps thanks for working again on this PR! Before reviewing, do you mind rebasing again on devel branch? Looks like the base is not up to date.

done!

fgyanz:
@marcosps  just noticed that in your PR `pylint` reported the following error: 
```
ERROR: Could not find a version that satisfies the requirement magic (from versions: none) 
ERROR: No matching distribution found for magic
```
If I'm not wrong, you need to modify the github CI script and install `python3-magic` package through zypper.
https://github.com/SUSE/klp-build/blob/main/.github/workflows/ci-tests.yml#L25

klpbuild/klplib/codestream.py
```diff
@@ -439,20 +439,34 @@ def get_mod_file_path(self, arch, mod):
 
         return Path(obj_match.group(1))
 
-    def find_obj_path(self, arch, mod):
-        # We already know the path to vmlinux, so return it
-        if not is_mod(mod):
-            return self.get_boot_file("vmlinux", arch).relative_to(get_datadir(arch))
+    def __search_obj_in_dir(self, fdir, fname):
+        file_path = list(fdir.glob(f"**/{fname}.*"))
```
vmezzela: given that we already know the correct `fdir` for a module, would `fdir.glob(f"{fname}.*")` just be enough?

tests/test_codestream.py
```diff
@@ -102,14 +102,14 @@ def test_get_mod_file_path():
 
 def test_find_obj_path():
     cs = Codestream("15.7u5", kernel="6.4.0-150700.53.19")
-    assert "lib/modules/6.4.0-150700.53.19-default/kernel/net/bluetooth/bluetooth.ko.zst" == str(
+    assert "lib/modules/6.4.0-150700.53.19-default/kernel/net/bluetooth/bluetooth.ko" == str(
```
vmezzela: Why this? Unless I did not understand the PR, I'd expect the *.zst version to be there, not the other way around.

marcosps: This happens when you deal with two different machines to test, and when you look at the same code for too much time :P

Let me fix this! Thanks for spotting this issue!

klpbuild/klplib/codestream.py
marcosps: I think so! Thanks!

PR #226, opened 2026-04-24
klplib/ibs: verify rpm download

fgyanz:
- When network is unstable or IBS unreliable, retrying the download might work.
- new `--verify-all` command: Verify that rpm packages have a valid md5sum (`rpm -K` ) and valid signature. 
- Delete corrupted cs and download them again next time `klp-build data --download` is executed.

### DIFF
--- a/klpbuild/klplib/data.py
+++ b/klpbuild/klplib/data.py
@@ -5,6 +5,7 @@
 
 import logging
 import shutil
+from pathlib import Path
 
 from klpbuild.klplib.utils import classify_codestreams_str,ARCHS
 from klpbuild.klplib.ibs import download_cs_rpms, verify_rpm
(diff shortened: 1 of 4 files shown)


marcosps (approved):
LGTM, thanks a lot!

fgyanz:
Hi guys, this one is ready for review.

klpbuild/klplib/data.py
```diff
@@ -107,5 +108,10 @@ def verify_all_rpms():
         logging.info("All RPMs passed verification")
 
 
+def __cs_is_missing_data(cs):
+    vmlinux = Path(cs.get_boot_dir(), cs.get_boot_filename("vmlinuz"))
```
marcosps: Is there some missing code that should call verify_rpm in this case?

marcosps (approved):
LGTM! Thanks!

klpbuild/klplib/data.py
fgyanz: Hi Marcos! Note that this code changed today. Mainly because this was only checking x86_64. A better approach is to check `vmlinux` on every arch. 

As for your question: If there's no `vmlinux`, then `--download` will attempt to extract the rpm if it exists (no rpm verification done), and if not, download it (verified on the fly). So no, at this point there's no verification. I can add it, as it does not hurt.

fgyanz: Done. Now, even if a rpm already exists, the --download will check if it's valid before extracting on `download_binary_rpms`

marcosps: Even better! Thanks!

vmezzela:
lgtm!

PR #227, opened 2026-04-24
klplib: ibs: fix find_missing_symbols

vmezzela:
find_obj_path was called with incorrect argument order and vmlinux_path was lacking the base directory.

### DIFF
--- a/klpbuild/klplib/ibs.py
+++ b/klpbuild/klplib/ibs.py
@@ -140,7 +140,7 @@ def get_cs_packages(cs_list, dest):
 
 
 def find_missing_symbols(cs, arch, lp_mod_path):
-    vmlinux_path = cs.find_obj_path("vmlinux", arch)
+    vmlinux_path = get_datadir(arch) / cs.find_obj_path(arch, "vmlinux")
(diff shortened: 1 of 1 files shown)


klpbuild/klplib/ibs.py
```diff
@@ -140,7 +140,7 @@ def get_cs_packages(cs_list, dest):
 
 
 def find_missing_symbols(cs, arch, lp_mod_path):
-    vmlinux_path = cs.find_obj_path("vmlinux", arch)
+    vmlinux_path = get_datadir(arch) / cs.find_obj_path(arch, "vmlinux")
```
marcosps, line 143: Ah! And you found the issue about the arguments being in the wrong disposition

marcosps (approved):
I was checking the calls to `find_obj_path`, and besides setup.py, the other two calls add `get_datadir` to it. The `find_obj_path` itself return a path relative to `get_datadir`. IMHO we should always keep the entire path returned by `find_obj_path` and call `relative_to` in setup.py alone.

What do you think?

fgyanz (approved):
LGTM! thanks a lot

PR #230, opened 2026-05-04
klplib/ibs: Do not verify downloaded rpm signature

fgyanz:
In order for the rpm verification to work the system is required to have SUSE certificates installed and up to date, which might not be the case in older servers.  In such cases the verification process will fail.
This commit removes the signature checking. The most important feature of the verification is not anti-tampering, but data corruption detection, which is still being done.


vmezzela (approved):
lgmt

marcosps:
LGTM!

PR #233, opened 2026-05-26
scan: Handle some bugzilla corner cases

fgyanz:


marcosps (commented):
LGTM!

vmezzela:
Hi @fgyanz, how do we distinguish from bug Fixed with a backport and bugs that were already fixed in base?

fgyanz:
@vmezzela From our side we can't really tell if a fix is actually in the base or not. However, we can trust the security-team for that. If they mark the parent bug as `RESOLVED-FIXED` *and* there are no backports for that CVE, then we can assume that the bug is somehow fixed.

Note that both conditions have to be true. If there's a backport available, klp-build will ignore the `RESOLVED-FIXED` and do the full check as usual.

vmezzela:
Just trying to find if we can miss some cases..

 What if the patch is in the product branches but it doesn't yet have the CVE reference to it? note that sometimes there's a delay (not small) before the reference is added to the patch. In this case, the bug **might** be closed as resolved-fixed and klp-build would find no backports for the respective CVE.

fgyanz:
> What if the patch is in the product branches but it doesn't yet have the CVE reference to it? note that sometimes there's a delay (not small) before the reference is added to the patch. In this case, the bug might be closed as resolved-fixed and klp-build would find no backports for the respective CVE.

Indeed, that can happen. And yes, that will show as `Fixed(0)` too. Which imho is correct. `Fixed(0)` means that security-team concluded that the vuln has been fixed somehow, but no backports were found. The Livepatch developer still has to manually check if that is actually true or not, usually by going to the parent bugzilla bug and reading if the backporter actually did something. One should never close a LP clone bug without first checking the parent bug.

Also, although a ref might be missing in a product branch, it's highly unlikely that it can happen in all affected product branches.

vmezzela:
> The Livepatch developer still has to manually check if that is actually true or not

So the question becomes: do we prefer to accept a false `Incomplete` instead of `Fixed(0)` (False Positive) when the commit is already in base, or a false `Fixed(0)` instead of `Fixed(n)` (False Negative) when the patch is just not updated with the CVE reference?

I think it's slightly easier to fall in the trap of the missing backport and assume we're covered, and therefore close the bug which should've been fixed. While on the other hand leaving an `Incomplete` entry in the CI does not have any actual bad consequence.

vmezzela:
On the other hand, it is also true that if the `Incomplete` that we have now is always wrong then the commit is already in base. While the `Fixed(0)` mentioned above only  **might** be wrong in some corner cases.

vmezzela:
Maybe we can add a note in the CI when we have `Fixed(0)` to make it clear that a double check is needed?

fgyanz:
> I think it's slightly easier to fall in the trap of the missing backport and assume we're covered, and therefore close the bug which should've been fixed. While on the other hand leaving an Incomplete entry in the CI does not have any actual bad consequence.

The `incomplete` section in the CI is insanely big, forever growing, and we can't check each of the bugs there. Thanks to this commit I was able to close 12 bugs, previously marked as `incomplete`, that were already fixed but without existing backports. We cannot leave already fixed bugs in the pending queue endlessly for several reasons. The most important one being that customers will get the wrong impression, as it happened in the past, and complain.

I understand your concerns, but I'm taking into account that the reviewer (anyone of us) will always have to check everything before closing a bug. This has always been the case. A bug cannot be closed without first manually checking that all codestreams are actually patched. And with the new `Fixed(0)` status, the reviewer will be warned beforehand that he'll be dealing with a `special` bug.

> Maybe we can add a note in the CI when we have Fixed(0) to make it clear that a double check is needed?

Yes, it can be done. But isn't the `Fixed(0)` status a clear sign that the bug is a corner case in some way or another? And as such, special care must be taken.

vmezzela:
> Yes, it can be done. But isn't the Fixed(0) status a clear sign that the bug is a corner case in some way or another? And as such, special care must be taken.


Right! Then LGTM! Thanks for the brainstorming! :)

PR #235, opened 2026-05-31
chore: remove explicit configparser dependency in setup.py

spaciouskarter78:
configparser is now a built-in utility into Python so there's no need to explicitly mention it as a dependency


marcosps:
LGTM

fgyanz (approved):
LGTM! Thanks a lot @spaciouskarter78 !

PR #237, opened 2026-06-09
bugzilla: Improve bug tracking and reporting

fgyanz:
- Add kss-watchdog support
- Add bug `deadline`, if available in `+-X` days format
- Change `priority` logic
- Minor code cleanup and robustness fixes

### DIFF
--- a/klpbuild/klplib/bugzilla.py
+++ b/klpbuild/klplib/bugzilla.py
@@ -159,7 +159,22 @@ def get_bug_subsys(bug):
 
 
 def get_bug_prio(bug):
-    return bug.priority[5:]
+    prio = bug.priority[5:]
+
+    '''
+    The bug priority has precedence over the severity.
+    However, "Medium" priority covers a wide range.
+    Add severity into the equation for more precision.
+    The severity rounds up or down the Medium priority.
+    '''
+    if prio == "Medium":
+        if bug.severity == "Critical":
+            prio = "High"
+        if bug.severity == "Normal":
+            prio = "Low"
+        # Severity "Major" is the most common for LPs.
+
+    return prio
 
 
 def is_bug_dropped(bug):
--- a/klpbuild/klplib/bugzilla.py
+++ b/klpbuild/klplib/bugzilla.py
@@ -195,12 +195,27 @@ def get_bug_prio(bug):
     return prio
 
 
+def get_kss_watchdog_report(bug):
+    comments = get_bug_comments(bug)
+
+    if not comments:
+        return ""
+
+    for c in comments:
+        if "kernel-security-sentinel" in c['creator']:
+            return c['text']
+
+    return ""
+
+
 def is_bug_dropped(bug):
-    return bug.resolution and bug.resolution in {"INVALID", "WONTFIX", "DUPLICATED"}
+    return (bug.resolution in {"INVALID", "WONTFIX", "DUPLICATED"} or
+            "NO CODESTREAM AFFECTED" in get_kss_watchdog_report(bug))
 
 
 def is_bug_fixed(bug):
-    return bug.resolution and bug.resolution == "FIXED"
+    return (bug.resolution == "FIXED" or
+            "NO ACTION NEEDED" in get_kss_watchdog_report(bug))
 
 
 def is_bug_embargoed(bug):
--- a/klpbuild/klplib/bugzilla.py
+++ b/klpbuild/klplib/bugzilla.py
@@ -100,7 +100,7 @@ def get_bug_comments(bug):
     while i < 5:
         try:
             return bug.getcomments()
-        except bugzilla.BugzillaError:
+        except Exception:
             # There's a max number of allowed simultaneous requests...
             time.sleep(5)
             i += 1
--- a/klpbuild/klplib/bugzilla.py
+++ b/klpbuild/klplib/bugzilla.py
@@ -5,14 +5,15 @@
 #          Marcos Paulo de Souza <mpdesouza@suse.com>
 
 from collections import namedtuple
+from datetime import date
 from functools import wraps
 import time
 import bugzilla
 
-from klpbuild.klplib.utils import is_cve_valid
+from klpbuild.klplib.utils import is_cve_valid, date_to_days
 from klpbuild.klplib.config import get_user_settings
 
-BugData = namedtuple("BugData", ["cve", "subsys", "cvss", "prio"])
+BugData = namedtuple("BugData", ["cve", "subsys", "cvss", "prio", "deadline"])
 
 __BZAPI = None
 def __check_is_connected(func):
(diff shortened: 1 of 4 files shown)


marcosps:
@fgyanz is this PR ready to be reviewed?

fgyanz:
No, not yet. I'm doing some changes affecting bugzilla. I'll batch several changes into this PR, given it's low priority.

fgyanz:
@marcosps it's ready for review :)

klpbuild/klplib/bugzilla.py
```diff
@@ -155,16 +160,61 @@ def get_bug_subsys(bug):
...
 def get_bug_prio(bug):
-    return bug.priority[5:]
+    prio = bug.priority[5:]
+
+    '''
+    The bug priority has precedence over the severity.
+    However, "Medium" priority covers a wide range.
+    Use the severity for more precision.
+    '''
+    if prio == "Medium":
```
vmezzela, line 188: what if the `prio <= Medium` and **Severity** is high? If that happens, then the `if prio == "Medium"` branch is not taken and the priority is not updated.

Do you want to achieve something like `prio = max(prio, severity)`?

klpbuild/klplib/bugzilla.py
```diff
@@ -155,16 +160,61 @@ def get_bug_subsys(bug):
...
+        if bug.severity == "Critical":
+            prio = "High"
+        if bug.severity == "Medium":
```
vmezzela: I cannot find severity = `Medium` in the bugzilla dropdown. Is that a mistake here?

klpbuild/klplib/bugzilla.py
```diff
@@ -155,16 +160,61 @@ def get_bug_subsys(bug):
...
+    '''
+    If there's no deadline for the LP, try with the parent bug.
+    This deadline is mainly meant for the backports, but still,
+    it might be useful for us. Filter out deadlines already past.
+    '''
+    if (dep := get_bug_dep(bug)) and hasattr(dep, "deadline") and dep.deadline:
+        if date.fromisoformat(dep.deadline) >= date.today():
+            return date_to_days(dep.deadline)
+
+    return "None"
```
vmezzela, line 176: Can't this just be `None` instead of a string? 

Not a problem for now, but if later in time we start doing `if BugData.deadline` somewhere, then this condition will evaluate to `True` even if there's no deadline.

klpbuild/plugins/scan.py
```diff
@@ -69,11 +82,11 @@ def scan_bugzilla():
         for b in bugs:
             if is_bug_embargoed(b):
                 continue
-            cve, system, cvss, prio  = get_bug_data(b)
-            if not cve:
+            data = get_bug_data(b)
+            if not data.cve:
```
vmezzela, line 86: What do you think about adding some logs here? just so we know that a bug without a CVE exists.

klpbuild/plugins/scan.py
```diff
@@ -84,42 +97,43 @@ def scan_bugzilla():
...
+    patches, _, _, affected_cs = scan(cve, None, None, False)
 
     npatches = len(set(f for _, files in patches.items() for f in files))
     if npatches:
-        status = f"Fixed({npatches})"
+        result.status = f"Fixed({npatches})"
 
     if dep and "security-team" not in dep.assigned_to:
-        status = f"Incomplete({npatches})"
+        result.status = f"Incomplete({npatches})"
```
vmezzela, line 126: I am confused: I thought the point of the commit `bugzilla: Use kss-watchdog to know bug status` was to rely on the kss-whatchdog to retrieve the status of a bug.

But if I understand correctly here when the bug has been marked completed by the kss-watchdog (i.e. `is_bug_fixed()` found the string `"NO ACTION NEEDED"`) but not yet assigned to the sec team bot, then this branch would mark the bug as `Incomplete`, overwriting the kss-watchdog result.

Or maybe I am making a huge confusion 😆

fgyanz: The kss-watchdog support is only partial and limited to bugs that do not require a fix. Those are:

- `NO ACTION NEEDED` because the fix is already in all branches. So we mark the bug as `FIXED(0)`. It has been fixed at some point in time. Then, klp-build runs `scan` to check if there are any backports with the CVE reference. If not, we keep the `FIXED(0)`.  Note that this doesn't mean a LP is not required (see #247).

- `CODESTREAMS NOT AFFECTED`  usually happens when a config/module is not supported. So no backports created, and `security-team` will eventually mark this on as `RESOLVED-WONT-FIX` or INVALID etc.


> But if I understand correctly here when the bug has been marked completed by the kss-watchdog (i.e. is_bug_fixed() found the string "NO ACTION NEEDED") but not yet assigned to the sec team bot, then this branch would mark the bug as Incomplete, overwriting the kss-watchdog result.

No, the kss-watchdog sets the assignee to `security-team` at the same time as it writes the comment.

klpbuild/plugins/scan.py
fgyanz: problem is that during a bugzilla scan we have all logs are disabled :/ 
We would need to write to a file instead. But I agree it's something we would need to have.

klpbuild/klplib/bugzilla.py
fgyanz: The "None" string is required for tabulate. Otherwise, passing a `None` will be shown as an empty space.

klpbuild/klplib/bugzilla.py
fgyanz: `prio <= Medium` You mean `prio < Medium`, right? Yes, it will be skipped. Reason being: A cloned LP bug is always created with `Medium` by default. If `security-team` sets priority to `Low`, that's intentional. And I've never seen a `Low` priority LP bug...

> Do you want to achieve something like prio = max(prio, severity)?

No, most LPs are `Medium` priority. Some exceptions are `High`. So if we want to have a more precise value we need to include `severity` into the equation. A very `Critical` severity bumps the priority, yes, but a low severity like `Normal`, decreases the priority. So, basically, we use the severity to round up or down the priority when it's unclear.

klpbuild/klplib/bugzilla.py
fgyanz: agh! good catch. Yes, my mistake while rebasing. SHould have been `Normal`. 

A `medium` priority + `normal` severity is below the average priority. So it becomes `Low`.

klpbuild/plugins/scan.py
fgyanz: actually, I can rise an ERROR log! better than nothing :)

klpbuild/plugins/scan.py
vmezzela: Great, thanks for the explanation @fgyanz !

fgyanz:
@vmezzela : Added new commit handling bugs with no CVE. Improved the priority logic explanation in the respective commit message. Fixed https://github.com/SUSE/klp-build/pull/237#discussion_r3519851547

vmezzela:
Thanks a lot!

PR #240, opened 2026-06-16
extract: drop __seg_gs macro after extraction

### LINKED ISSUE #234: Drop `__seg_gs` empty definition
When extracting livepatches for SLE16, we add " -D__seg_gs=" to the klp-ccp compiler command line. This results in klp-ccp emitting an empty `#define __seg_gs`.

Drop it from the extracted code in since it can cause issues when extracting code that uses named address spaces.

I believe that `lp_out_cleanup` is a good place to look into the issue.

vmezzela:
Fixes: #234

This PR intentionally targets `main`, we can back-merge this into devel afterward.

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -32,7 +32,8 @@ UNSUPPORTED_MACROS = [
         "__KERNEL__",
         "MODULE",
         "KBUILD_MODNAME",
-        r"__has_attribute\("
+        r"__has_attribute\(",
+        "__seg_gs"
         ]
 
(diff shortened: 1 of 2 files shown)


fgyanz (changes requested):
Looks good! Do you mind updating `test_lp_out_cleanup_removes_unsupported_macros` with the new macros?

vmezzela:
Right @fgyanz, done! :heavy_check_mark:

fgyanz (approved):
LGTM! thanks :)

marcosps (approved):
LGTM!

PR #241, opened 2026-06-16
Prepare cleanup

marcosps:
These patches are mostly cosmetic but the last one. It reduced the time to run prepare tests for me from 25 minutes to 12 minutes, which itself is huge.  I'm using workers=4, but the gains could be much higher if the number of workers is bigger.

### DIFF
--- a/klpbuild/plugins/prepare_tests.py
+++ b/klpbuild/plugins/prepare_tests.py
@@ -75,7 +75,7 @@ def download_built_rpms(lp_name, lp_filter):
     logging.info("Download finished.")
 
 
-def prepare_tests(lp_name, lp_filter):
+def run(lp_name, lp_filter):
     test_src = get_tests_path(lp_name)
     if test_src and not os.access(test_src, os.X_OK):
         logging.error("Script %s has no execution bit set. Aborting", test_src)
@@ -153,7 +153,3 @@ def prepare_tests(lp_name, lp_filter):
         )
 
         logging.info("Done.")
-
-
-def run(lp_name, lp_filter):
-    prepare_tests(lp_name, lp_filter)
--- a/klpbuild/plugins/prepare_tests.py
+++ b/klpbuild/plugins/prepare_tests.py
@@ -114,15 +114,12 @@ def prepare_tests(lp_name, lp_filter):
                 logging.info("%s/%s: rpm dir not found. Skipping.", cs.full_cs_name(), arch)
                 continue
 
-            # TODO: there will be only one rpm, format it directly
+            rpms = list(rpm_dir.glob("**/*.rpm"))
+            assert len(rpms) == 1, f"{cs.full_cs_name()}/{arch}: {len(rpms)} rpms found. Expecting only one"
 
-            rpms = [rpm for rpm in os.listdir(rpm_dir) if rpm.endswith(".rpm")]
-            if len(rpms) > 1:
-                raise RuntimeError(f"ERROR: {cs.full_cs_name()}/{arch}. {len(rpms)} rpms found. Excepting to find only one")
-
-            for rpm in rpms:
-                validate_livepatch_module(cs, arch, rpm_dir, rpm)
-                shutil.copy(Path(rpm_dir, rpm), Path(test_arch_path, "built"))
+            rpm_file = rpms[0]
+            validate_livepatch_module(cs, arch, rpm_dir, rpm_file)
+            shutil.copy(Path(rpm_dir, rpm_file), Path(test_arch_path, "built"))
 
             if cs.rt and arch != "x86_64":
                 continue
--- a/klpbuild/plugins/prepare_tests.py
+++ b/klpbuild/plugins/prepare_tests.py
@@ -121,9 +121,6 @@ def prepare_tests(lp_name, lp_filter):
             validate_livepatch_module(cs, arch, rpm_dir, rpm_file)
             shutil.copy(Path(rpm_dir, rpm_file), Path(test_arch_path, "built"))
 
-            if cs.rt and arch != "x86_64":
-                continue
-
             build_cs.append(cs.get_full_product_name())
 
         logging.info("Done.")
--- a/klpbuild/plugins/prepare_tests.py
+++ b/klpbuild/plugins/prepare_tests.py
@@ -3,20 +3,35 @@
 # Copyright (C) 2025 SUSE
 # Author: Marcos Paulo de Souza <mpdesouza@suse.com>
 
+import concurrent.futures
 import importlib
 import logging
 import os
-from pathlib import Path
 import shutil
 import subprocess
 import sys
+from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


klpbuild/plugins/prepare_tests.py
```diff
@@ -75,7 +90,22 @@ def download_built_rpms(lp_name, lp_filter):
...
-def prepare_tests(lp_name, lp_filter):
+def validate_module_and_copy(cs, arch, lp_name, test_arch_path):
+    """Validate a livepatch module and copy it to the test directory."""
+    rpm_dir = Path(cs.get_ccp_dir(lp_name), arch, "rpm")
+    if not rpm_dir.exists():
+        logging.info("%s/%s: rpm dir not found. Skipping.", cs.full_cs_name(), arch)
+        return
+
+    rpms = list(rpm_dir.glob("**/*.rpm"))
+    assert len(rpms) == 1, f"{cs.full_cs_name()}/{arch}: {len(rpms)} rpms found. Expecting only one"
```
fgyanz: I'd be very cautious when using asserts. Mainly because if the reason that triggered the assert is reproducible it will block testing any other codestream.

 I would rather not abort the whole `prepare-tests` operation just for a missing rpm file. A warning would be better imho. What do you think?

klpbuild/plugins/prepare_tests.py
```diff
@@ -75,7 +90,22 @@ def download_built_rpms(lp_name, lp_filter):
...
+
+    rpm_file = rpms[0]
+    validate_livepatch_module(cs, arch, rpm_dir, rpm_file)
+    shutil.copy(Path(rpm_dir, rpm_file), Path(test_arch_path, "built"))
```
fgyanz: shouldn't we do a move instead?

fgyanz (commented):
Very nice @marcosps ! Thanks a lot for the PR. I commented some concerns and nitpicks. But looks great otherwise.

klpbuild/plugins/prepare_tests.py
marcosps: Well, it used to happen in the past when the directory containing the rpms wasn't being removed. Maybe I can even drop this assert, since if it does find duplicated rpms it means we have a bigger issue.

klpbuild/plugins/prepare_tests.py
marcosps: I think so, yes. let me change it.

marcosps:
@fgyanz I resolved your comments, and retested, and everything works as intended!

klpbuild/plugins/prepare_tests.py
```diff
@@ -75,7 +90,20 @@ def download_built_rpms(lp_name, lp_filter):
...
-def prepare_tests(lp_name, lp_filter):
+def validate_module_and_move(cs, arch, lp_name, test_arch_path):
+    """Validate a livepatch module and move it to the test directory."""
+    rpm_dir = Path(cs.get_ccp_dir(lp_name), arch, "rpm")
+    if not rpm_dir.exists():
+        logging.info("%s/%s: rpm dir not found. Skipping.", cs.full_cs_name(), arch)
+        return
+
+    # There should be only one rpm, so take the first entry in the list
+    rpm_file = list(rpm_dir.rglob("*.rpm"))[0]
```
fgyanz: Sorry for insisting, but wouldn't this crash the whole process if theres no rpm for a specific cs? Perhaps we can handle the case when len() is 0, `warn` and return.

marcosps: Well, it never happened for quite a long time, but I can of course add a warn here. But I don't think it's needed :)

fgyanz: Ok :) It's just that I'm worried that missing a single rpm might lead to blocking someone from testing a LP. Perhaps an unreasonable fear...
If you think it's not needed then all good

marcosps: I believe that I finally solved the bug that I mentioned in the last commit. Fortunately I just hit it, so now we are safe!

klpbuild/plugins/prepare_tests.py
```diff
@@ -75,6 +90,19 @@ def download_built_rpms(lp_name, lp_filter):
...
 
+def validate_module_and_move(cs, arch, lp_name, test_arch_path):
+    """Validate a livepatch module and move it to the test directory."""
+    rpm_dir = Path(cs.get_ccp_dir(lp_name), arch, "rpm")
+    if not rpm_dir.exists():
+        logging.info("%s/%s: rpm dir not found. Skipping.", cs.full_cs_name(), arch)
+        return
+
+    # There should be only one rpm, so take the first entry in the list
+    rpm_file = list(rpm_dir.rglob("*.rpm"))[0]
```
vmezzela: What if there's no rpm here?

I understand that @fgyanz suggested to drop the assert and avoid to bail out in this case, but we still need to perform the error handling: otherwise with the current code we'd get `IndexError: list index out of range` and the process crashes, which is the same as having the previous assert.

The error handling should also skip the call to `validate_livepatch_module()` and shouldn't even add the codestream to `build_cs`, right?

vmezzela (changes requested):

klpbuild/plugins/prepare_tests.py
marcosps: But if we have not rpms at this point, it means that we have an issue in IBS, so it's better to let it explode so we can find this issue and fix it. We have had some issues in the code, specially when the codestreams had the wrong patchinfoid, so it's better for us to find it and fix.

vmezzela: > But if we have not rpms at this point, it means that we have an issue in IBS

Then you could have left the previous `assert()`, so the user at least gets a clear error message instead of a random `IndexError` that need to be deciphered.

IIUC @fgyanz was asking not to crash the process if that happens, and I agree with him for the reasons he mentioned: It's not worth to block someone from testing the livepatch just because there's a problem somewhere else (IBS in this case).

> so it's better to let it explode so we can find this issue and fix it.

Sorry, I do not agree. If you emit a warning message then the user can still see that there's an issue to be solved, but without the extra cost of not being able to test the livepatch.

Also, since we're now packaging klp-build, I would opt for a more fault-tolerant approach. I’d prefer the tool to handle issues gracefully and keep working, rather than bailing or crashing the entire process whenever it hits an error.

If it crashes also on minor issues like this one, then the user is always forced to solve it before it can continue to work on the livepatch. I am pretty sure this is bad, especially given that this kind of issues tend to happen when there are urgent livepatch to work on (because of the Murpy's Law :joy: ).

marcosps:
Ok, I can add a check and skip the given codestream if there are no rpms or more than one.

fgyanz (approved):
LGTM! thanks :)

vmezzela:
lgtm! Thanks for addressing the comments!

PR #242, opened 2026-06-17
klplib/affected_file: introduce data structures for affected files/module/config

vmezzela:


marcosps (commented):
Maybe the last commit would be better placed as second commit?

vmezzela:
> Maybe the last commit would be better placed as second commit?

I think I will drop the commit altogether because I found another workaround for the issue that such commit tried to solve, and it was actually buggy..

klpbuild/klplib/analysis.py
```diff
@@ -244,12 +252,17 @@ def filter_unsupported_kmodules(cs_list):
...
             unset_cs.append(cs)
 
-        # Cleanup for future re-use
-        cs.modules.clear()
+        # NOTE: previously cs.modules.clear() was called here to wipe the
+        # supportedness bools so the path-cache repopulation in
+        # setup.__setup_check_mod could reuse the dict. With AffectedModule the
+        # supported/blacklisted flags and the per-arch obj-path cache are
+        # independent fields on the same object, so the wipe is no longer
+        # needed - and would in fact destroy state we want to keep.
```
marcosps: This comment in interesting, along with other one above to explain how things used to be (before this patch), but I believe that people could use git to see what changed if so needed. I believe that these commends can be removed.

marcosps (approved):
I liked the changes. IIUC, the codestreams.json format is the same, we are just using classes to separate the entries instead of using plain dicts, which is much better.

LGTM!

klpbuild/klplib/analysis.py
vmezzela: Right, I will drop them

fgyanz:
> the codestreams.json format is the same

@marcosps 
A word on this. The json is not the same:
```
             "files": {
                 "fs/smb/client/smb2ops.c": {
-                    "symbols": [
+                    "config_name": "CONFIG_CIFS",
+                    "module_name": "fs/smb/client/cifs",
+                    "affected_symbols": [
                         "crypt_message"
                     ],
-                    "conf": "CONFIG_CIFS",
-                    "module": "fs/smb/client/cifs",
                     "ibt": true,
+                    "dup_symbols": [],
                     "ext_symbols": {
                         "cifs": [
                             "__SCK__tp_func_smb3_ses_not_found",
@@ -167,7 +170,20 @@
                 }
             },
             "modules": {
-                "fs/smb/client/cifs": "usr/lib/modules/6.4.0-33-rt/kernel/fs/smb/client/cifs.ko.zst"
+                "vmlinux": {
+                    "supported": true,
+                    "blacklisted": false,
+                    "obj_paths": {
+                        "x86_64": "usr/lib/modules/6.4.0-33-rt/vmlinux.xz"
+                    }
+                },
+                "fs/smb/client/cifs": {
+                    "supported": true,
+                    "blacklisted": false,
+                    "obj_paths": {
+                        "x86_64": "usr/lib/modules/6.4.0-33-rt/kernel/fs/smb/client/cifs.ko.zst"
+                    }
+                }
             },
```
So this PR effectively breaks `codestreams.json` for older klp-build versions. Nothing wrong with it, just a fact. 
@vmezzela Thanks a lot for the change. It looks great! I'm currently testing it. I'll report any unexpected behavior.

marcosps:
Ok, I see. I ran the changes locally and I liked the changes. They are not so big, and makes things more clear about the uses of each field. Thanks a lot @vmezzela !

vmezzela:
Addressed https://github.com/SUSE/klp-build/pull/242#discussion_r3499075243 .

klpbuild/klplib/analysis.py
```diff
@@ -244,13 +247,11 @@ def filter_unsupported_kmodules(cs_list):
         if not cs.modules:
             continue
 
-        supported = [s for _, s in cs.modules.items() if s]
-        if not supported:
+        # A module is "supported" iff its klp-build supportedness flag is True
+        # (None / False are not).
+        if not any(m.supported for m in cs.modules.values()):
```
fgyanz, line 252: It seems this is not working as expected. Running `klp-build setup -n bsc1266727`:
```
[*] Analysing required CONFIGs...

6.0rtu9-24 15.7rtu1-15 16.0rtu0-13:
CONF: CONFIG_RDMA_RXE
x86_64: m
ppc64le: n
s390x: n

6.0u8-23:
CONF: CONFIG_RDMA_RXE
x86_64: m
ppc64le: n
s390x: m

12.5u69-83 15.4u41-54 15.5u27-40 15.6u12-26 15.7u1-15:
CONF: CONFIG_RDMA_RXE
x86_64: m
ppc64le: m
s390x: n

16.0u0-13:
CONF: CONFIG_RDMA_RXE
x86_64: m
ppc64le: m
s390x: m

[*] Analysing affected kernel modules...

6.0rtu9-24 6.0u8-23 12.5u69-83 15.4u41-54 15.5u27-40 15.6u12-26 15.7rtu1-15 15.7u1-15:
MOD: drivers/infiniband/sw/rxe/rdma_rxe
Supported: True

16.0rtu0-13 16.0u0-13:
MOD: drivers/infiniband/sw/rxe/rdma_rxe
Supported: False

Affected architectures:
	ppc64le s390x x86_64
	6.0rtu25 6.0u24-25 12.5u84 15.4u55-56 15.5u41-42 15.6u27-28 15.7rtu16-17 15.7u16-18 16.0rtu14 16.0u14
All affected codestreams:
	6.0rtu9-24 6.0u8-23 12.5u69-83 15.4u41-54 15.5u27-40 15.6u12-26 15.7rtu1-15 15.7u1-15 16.0rtu0-13 16.0u0-13
```

Codestreams `16.0rtu0-13 16.0u0-13` with unsupported modules are not being marked as unaffected. Could it because vmlinux is set by default in cs.modules?

fgyanz (changes requested):
Hi @vmezzela ! Tested and reviewed. Only one possible bug found. Everything  else looks great! Good job :)

klpbuild/klplib/analysis.py
vmezzela: Hi @fgyanz , thanks for spotting this. This is a regression introduced by this PR: the function `find_obj_path` has a caching behavior, and was used by the data-availability probe `__cs_is_missing_data()` that checks for the vmlinux presence and as a side effect added vmlinux to the cs.modules. `AffectedModule.vmlinux()` is always supported and therefore the `filter_unsupported_kmodules()` doesn't filter the codestream out.

Will fix this soon.

marcosps: I tested the latest version, so now it's fixed. Thanks a lot!

klpbuild/klplib/file2config.py
```diff
@@ -216,6 +216,15 @@ def find_file_config(cs, path):
...
         return '', ''
 
+    # If the config is built into the kernel image (=y) on every arch where
+    # it's set, the object lives in vmlinux. Report vmlinux directly so the
+    # stored module_name matches what get_file_mod() resolves at runtime (and
+    # what the manual setup path defaults to).
+    cfg = cs.get_all_configs(config)
+    if cfg.is_set() and all(cfg.get_arch(arch) is ConfigState.BUILTIN
+                            for arch in cfg.archs()):
+        obj = AffectedModule.VMLINUX
```
fgyanz, line 225: I have mixed feelings about this change. IMHO the function should return the object indicated in the makefile (so, *.o). The above heuristic could be done at some higher level, no? Perhaps even on the fly when needed. It could get kind of confusing what to expect from the function otherwise.

klpbuild/klplib/codestream.py
```diff
@@ -520,20 +520,25 @@ def __search_obj_in_dir(self, fdir, fname):
...
+        ``"fs/ext4/ext4"`` or ``"vmlinux"``) on ``arch``.
+
+        The resolved path is cached per-arch on the matching
+        :class:`AffectedModule` in :attr:`modules` *only when one already
+        exists*. This method must never create a module entry: resolving or
+        probing a path (the data-availability check in ``data.py``, the ibs
+        vmlinux lookup, symbol checks) must not register a module as
+        "affected", otherwise :func:`filter_unsupported_kmodules` would be
+        defeated by a spurious ``supported=True`` vmlinux entry landing in
+        every codestream's :attr:`modules`.
```
fgyanz, line 526: a bit verbose, no? Do we need that much detail?

klpbuild/klplib/analysis.py
fgyanz: Thanks @vmezzela ! Tested and works as expected :)

fgyanz (changes requested):
Just a few remarks about the new commits added in the latest update. Otherwise everything looks good :)

klpbuild/klplib/file2config.py
vmezzela: I added this commit to workaround one issue that was present even before this PR, that is: even when the module is builtin, the codestream object contained the path of a *.ko (instead of vmlinux), even if that object did not exists. The current refactor just made the issue more clear and Marcos noticed it by reading the codestreams.json file.

What do you suggest to workaround this? It does not make sense to return a path to any object when the module is builtin, right?

klpbuild/klplib/codestream.py
vmezzela: otherwise it's not obvious why we're not always caching the object: we could end up doing the same mistake again

klpbuild/klplib/file2config.py
vmezzela: I also dislike the fact that this commit make file2config depending on the cs object rather than only on the plain Makefiles.

> What do you suggest to workaround this? It does not make sense to return a path to any object when the module is builtin, right?

If I remember correctly, this thing was not even causing any issue. It was just about logical correctness of the `codestream.jsn` file. But maybe @marcosps remembers better

marcosps: I don't recall it being different. TBH I just noticed it last week. Things have been changing so fast that I don't recall, but in the past, long past, we used to point to the object file (a .ko or vmlinux), but I'm not sure about recently changes.

fgyanz: > What do you suggest to workaround this? It does not make sense to return a path to any object when the module is builtin, right?

afaik the function returns the object (*.o) **before** linking to a ko or vmlinux. But the *.o exists in both cases, and that's what we fetch from the Makefile.

> If I remember correctly, this thing was not even causing any issue. It was just about logical correctness of the codestream.jsn file.

Maybe we can find a better way to display this information in the json and/or stdout?

marcosps: I was expecting the mapping between
files
  files
     `module_name`
modules
  <module_name>
  
  With at least some resemblance. The `files:module_name` could be a path to a module, but it's not useful to have the name of the compiled .o file, IMHO. It should be the path to the module being livepatched, or vmlinux.
  
  What do you think about it? In the past it used to be like this, but I'm not sure how long ago it was...

vmezzela: > > What do you suggest to workaround this? It does not make sense to return a path to any object when the module is builtin, right?
> 
> afaik the function returns the object (*.o) **before** linking to a ko or vmlinux. But the *.o exists in both cases, and that's what we fetch from the Makefile.

Yes, and that is what I wanted to change. After looking at the code again I still cannot understand why we should return it.

> With at least some resemblance. The files:module_name could be a path to a module, but it's not useful to have the name of the compiled .o file, IMHO. It should be the path to the module being livepatched, or vmlinux.

I agree with @marcosps

vmezzela: @fgyanz another issue that this commit was fixing is that the affected module was changing between the automation and the manual setup (with `--conf`, `--file-funcs`).

When a file was builtin, the reported affected obj was `fs/notify/mark` with the automation and `vmlinux` with the manual setup.

This commit makes the two consistent, as they should be.

PR #243, opened 2026-06-18
klplib: ibs: Make delete_built_rpms delete files using ARCHS

marcosps:
The meaning of cs.archs has been changing from time to time. In order to successfully delete the rpms downloaded to all available architecures that we support, use ARCHS instead of cs.archs.


marcosps:
I'll add this commit into my current PR that reworks prepare-tests

PR #244, opened 2026-06-23
klplib: codestream: set default value for 'eol' and 'required_patches'

ldzhong:
it's possible to get a key error if the file codestreams.json doesn't have the 'eol' and 'required_patches' items, which was generated by an old version klp-build tool.
===============================
While in the process of working for a LP project, I updated the klp-build tool. The original codestreams.josn doesn't 
include 'eol' and 'required_patches' items, so I get the following error:
```
# klp-build cs-diff -n bsc1262759 --filter '16.0u1|16.0u9'
  File "/root/.local/bin/klp-build", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/work/source/klp-build/klpbuild/main.py", line 24, in main
    load_codestreams(args.lp_name)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/work/source/klp-build/klpbuild/klplib/codestreams_data.py", line 48, in load_codestreams
    _codestreams[cs] = Codestream.from_data(json_cs[cs])
                       ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/work/source/klp-build/klpbuild/klplib/codestream.py", line 68, in from_data
    data["kernel"], data["eol"], data["archs"], data["files"],
                    ~~~~^^^^^^^
KeyError: 'eol'
```

The same error for key 'required_patches' .

### DIFF
--- a/klpbuild/klplib/codestream.py
+++ b/klpbuild/klplib/codestream.py
@@ -61,6 +61,9 @@ class Codestream:
 
     @classmethod
     def from_data(cls, data):
+        data.setdefault("eol", "")
+        data.setdefault("required_patches", None)
+
         return cls(data["name"], data["project"], data["patchid"],
                    data["kernel"], data["eol"], data["archs"], data["files"],
                    data["modules"], data["configs"], data["required_patches"])


vmezzela:
Hi Lidong, I am not sure this works correctly across all the steps to create a livepatch. I think for instance the `extract` command now requires the `required_patches` to be set so it know what patches to apply before the extraction.

I think the correct approach is to update klp-build only when the current livepatch is complete and you start working on a new one, so the setup runs from scratch.

ldzhong:
> Hi Lidong, I am not sure this works correctly across all the steps to create a livepatch. I think for instance the `extract` command now requires the `required_patches` to be set so it know what patches to apply before the extraction.
> 
> I think the correct approach is to update klp-build only when the current livepatch is complete and you start working on a new one, so the setup runs from scratch.

Hi Vincenzo, thanks for your reply. Yes it's kind of a corner case that I got when I updated the klp-build package while working for one LP project. If I use the same klp-build I guess this error won't be triggered. Feel free to close the PR if you think it doesn't make much sense.

vmezzela:
@fgyanz @marcosps do you have different opinions before I close the PR?

fgyanz:
@vmezzela nope, I agree :+1: 

Thanks anyway @ldzhong ! But as Vincenzo said, for simplicity we assume with a new release a clean state. So, no backward compatibility with the per livepatch data.

marcosps:
We faced this issue before. From time to time we update the json format to include/remove new data, and we expect that a livepatch is created and finished using the same version to avoid such issues.

PR #245, opened 2026-06-23
Packaging

vmezzela:


marcosps (approved):
LGTM.

It would be nice to check how does src.opensuse.org can be used to trigger builds on OBS, if it's not being used in this version. Either way, it's nice to have the .spec file here.

Thanks!

vmezzela:
imho the best thing would be to let github trigger an automatic build in OBS whenever we update the main branch, but I still haven't managed to achieve this behavior

fgyanz (approved):
LGTM! thanks! Just a small detail: do you mind updating the README.md to show that klp-build can installed with zypper?

vmezzela:
> LGTM! thanks! Just a small detail: do you mind updating the README.md to show that klp-build can installed with zypper?

done! :heavy_check_mark:

PR #248, opened 2026-07-06
Fix --conf-mod-file-funcs argument of setup subcommand

marcosps:
This work was based on top of #242 . This is needed when we need to create a livepatch targetting two different subsystems  with different modules and configs, usually meaning that we can't really use the upstream fix and we need a manual approach.

Please review!

### DIFF
--- a/klpbuild/klplib/affected_file.py
+++ b/klpbuild/klplib/affected_file.py
@@ -1,0 +1,276 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2026 SUSE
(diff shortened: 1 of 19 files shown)


vmezzela:
lgtm! Let's wait the other PR to get reviewed and merged and then we can merge this if @fgyanz approves :)

fgyanz:
lgtm! This issue has been annoying me for a long time. Thanks a lot for fixing it!

marcosps:
I found some issues regarding `ifdefs` being added when they are not needed for two codestreams of the same type (-RT). While this fix makes the code to no crash, it still does not work the same way of the current `--file-funcs`.

PR #249, opened 2026-07-13
plugins: extract: Remove __sched attribute

marcosps:
The __sched is translated to __section(".sched.text"), which isn't handled by the module linker script. The functions inside this section are not shown in the stack traces, but it's only meaningful for functions builtin in the kernel, and not for modules, such as livepatches.


vmezzela (approved):
lgtm

PR #250, opened 2026-07-23
Check parent bug

marcosps:
There are times that a bug is under embargo, or don't have a CVE assigned, but these patches have references to the bug number that is tracking the vulnerability. Use the parent bug number to help getting these patches.

Created on top of Vincenzo's PR.

### DIFF
--- a/klpbuild/klplib/ksrc.py
+++ b/klpbuild/klplib/ksrc.py
@@ -190,18 +190,22 @@ def store_patch(pfile, patch, savedir, savedir_idx, bc):
         f.write(pfile)
 
 
-def get_branch_patches(cve, mbranch):
+def get_branch_patches(cve, mbranch, parent=None):
     kern_src = get_user_path('kernel_src_dir')
+
+    search_string = f"CVE-{cve}"
+    if parent:
+        search_string += f"|bsc#{parent}"
 
     try:
         patch_files = subprocess.check_output(
-            ["/usr/bin/git", "-C", kern_src, "grep", "-l", f"CVE-{cve}",
+            ["/usr/bin/git", "-C", kern_src, "grep", "-E", "-l", search_string,
              f"remotes/origin/{mbranch}", "--", "patches.suse/"],
             stderr=subprocess.STDOUT,
         ).decode(sys.stdout.encoding)
     except subprocess.CalledProcessError:
         # If we don't find any commits for RT branchs, try with the non-RT variant.
-        return [] if "RT" not in mbranch else get_branch_patches(cve, mbranch.replace("-RT", ""))
+        return [] if "RT" not in mbranch else get_branch_patches(cve, mbranch.replace("-RT", ""), parent)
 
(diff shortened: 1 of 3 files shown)


vmezzela:
I like it. However, I still have to make sure #242 is ready to be merged. After that's merged I will review it more carefully. :)

marcosps:
> I like it. However, I still have to make sure #242 is ready to be merged. After that's merged I will review it more carefully. :)

Sure, I agree! Just wanted to share the PR so you could also use it if your current LP doesn't have the CVE tag added.

klpbuild/klplib/ksrc.py
```diff
@@ -261,7 +269,7 @@ def get_patches(cve, savedir=None, extra_patches=None):
             else:
                 logging.debug("\textra patch %s not found in %s", extra_patch, mbranch)
 
-        for patch in get_branch_patches(cve, mbranch):
+        for patch in get_branch_patches(cve, mbranch, parent):
```
fgyanz, line 268: I would propose renaming `parent` to `bsc`. In the `ksrc.py` context, `parent` doesn't mean much. We are tracking backports for a specific CVE and bsc.

In `setup` and `scan` it makes sense to have `'parent` though :)

klpbuild/klplib/ksrc.py
```diff
@@ -113,6 +113,10 @@ def get_patch_kernel_commit(patch, branch):
         # skip it, and try with the next found kernel commit.
         # This is a very corner case, so try at most 3 times.
         _, _, suse_commit = get_commit_data(kernel_commit)
+        # TODO: check why some commits on kernel expanded tree does not contain suse-commit
```
fgyanz: Did you find a commit without the suse-commit ? In theory, all commits introducing a backport from kernel-source **must** include a suse-commit tag.

klpbuild/plugins/setup.py
```diff
@@ -178,8 +178,13 @@ def setup_codestreams(lp_name, data):
         all_codestreams = get_supported_codestreams()
         codestreams = utils.filter_codestreams(data["lp_filter"], all_codestreams)
     else:
+        bug = bugzilla.get_bug(lp_name)
+        if len(bug.depends_on) == 1:
+            data["parent"] = bug.depends_on[0]
+            logging.info("Using parent bsc number to help tracking backports: %s", data["parent"])
```
fgyanz, line 184: Maybe we should do this too when running `scan` command? If so, we can have this in scan() rather than in setup().

fgyanz (changes requested):

klpbuild/plugins/setup.py
marcosps: First I tried to isolate the behavior on setup, but I agree that scan should be using it as well. Also, there was that issue with missing `suse-commit` info that I would like to understand. I posted the PR so we could discuss if this is interesting for us, and I'll try to iron it out this week and publish a new version then #242 is merged.

Issue #254, opened 2026-08-03
Failed to find klpp function during extract


alisarctl:
klp-build setup -n bsc1269037 --conf CONFIG_IP_SCTP --module sctp --file-funcs net/sctp/sm_statefuns.c sctp_sf_cookie_echoed_err

The function in question is 

`enum sctp_disposition klpp_sctp_sf_cookie_echoed_err(...`

at the moment of the extract, Failed to find klpp_sctp_sf_cookie_echoed_err in `work_net_sctp_sm_statefuns.c/bsc1269037_net_sctp_sm_statefuns.c`

And thus the header file doesn't have any generated forward declarations.

fgyanz:
Fixed in #257. Thanks for reporting the bug, @alisarctl .

Issue #253, opened 2026-08-03
Forward declaration sometimes missing last letter in struct name


alisarctl:
For example when klp-build generates forward declarations for the following function

`int ipc_addid(struct ipc_ids *, struct kern_ipc_perm *, int);`

it writes
`struct sctp_chun;`
`struct kern_ipc_per;`

instead of 
`struct sctp_chunk;`
`struct kern_ipc_perm;`

fgyanz:
Fixed in #257. Thanks for reporting the bug, @alisarctl .

PR #257, opened 2026-08-05
extract: Support non-static klpp symbols

fgyanz:
Fix the regex so that it also extracts non-static klpp funtion signatures.
And skip klpp function declaration when looking for prototypes.
Fixes: #254 and #253

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -330,7 +330,7 @@ def get_klpp_symbols(out_dir, lp_out):
             sym = sym.strip()
             # Create a regex for finding the klpp_{sym} function
             # declaration and definition.
-            rfmt = fr"(static\s+)(([\w\*]\s*)*klpp_{sym}\s*\((?:[^()]*|\([^()]*\))*\))"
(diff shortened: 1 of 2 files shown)


klpbuild/plugins/extract.py
```diff
@@ -330,7 +330,7 @@ def get_klpp_symbols(out_dir, lp_out):
             sym = sym.strip()
             # Create a regex for finding the klpp_{sym} function
             # declaration and definition.
```
vmezzela: The comment is outdated

vmezzela (changes requested):
besides the comment I left inline, it lgtm!

marcosps:
These regexes are monstrous, but I believe that we don't have other choices right now... LGTM!

fgyanz:
> These regexes are monstrous, but I believe that we don't have other choices right now... LGTM!

Indeed... but unless we build some kind of AST like klp-ccp, I don't see any other way of generating the prototypes. Perhaps we should tweak `klp-ccp` to output the functions prototype along with the klpp symbols in `patched_funcs`.

PR #258, opened 2026-08-05
klplib: file2config: Return empty config is the file doesn't exists

marcosps:


vmezzela:
What is this fixing?

I haven't looked deeply at the code but note that:
- this change makes `_find_config_include` unreachable
- there's a typo in the commit subject

We already have something similar some lines below:

```py
    if not config or not config.startswith('CONFIG_'):
        return '', ''
```

I think it's just a matter of reorganizing the code and move this above but without overshadowing `_find_config_include` ?

marcosps:
IIUC, the issue is that the file doens't exists, so the config returned is empty.

marcosps:
@vmezzela agreed, I just reordered the code as you suggested.

fgyanz:
The change still makes `_find_config_include` unreachable, no?

vmezzela:
> The change still makes `_find_config_include` unreachable, no?

Exact!

@marcosps I suggested something:

```py
    if not config:
        config, obj = _find_config_include(cs, obj_file.parent, obj_file.name)

    # Return empty if the file doesn't exists
    if not config or not config.startswith('CONFIG_'):
        return '', ''

    cfg = cs.get_all_configs(config)
```

But I do not know if this reordering can have unexpected side effects. I neither looked at it carefully nor tested it yet.

vmezzela:
@marcosps nvmd I found the issue and pushed here :)

fgyanz (approved):
lgtm! thanks :)

Would it make sense to also have this?
```
class AffectedConfig:
         ...
        if not config_name or not config_name.startswith("CONFIG_"):
            raise ValueError(
                f"Invalid config '{config_name}'")
```

Though it should not happen, it doesn't harm either way given that we already check for valid config. What do you think?

marcosps:
In the case of the file is part of `arch`, but it don't return any values in that elif, we then call get_all_configs twice? Not that it does any impact, just wondering if we could avoid it. Besides that, LGTM. Thanks a lot for fixing it!

After this fix is merged, we could try running checking if #259 (and all the other PRs) runs without breaking any other tests. Thanks!

vmezzela:
> lgtm! thanks :)
> 
> Would it make sense to also have this?
> 
> ```
> class AffectedConfig:
>          ...
>         if not config_name or not config_name.startswith("CONFIG_"):
>             raise ValueError(
>                 f"Invalid config '{config_name}'")
> ```
> 
> Though it should not happen, it doesn't harm either way given that we already check for valid config. What do you think?

we already have:

```py
    def __init__(self, config_name: str,
                 arch_values: dict[str, str] | None = None):
        if not config_name.startswith("CONFIG_"):
            raise ValueError(
                f"Invalid config '{config_name}': missing CONFIG_ prefix")

```

Do you mean addinf the `if not config_name ...`?

fgyanz:
Yep. It's just as suggestion, the main issue has already been fixed and we can merge if you think it's not needed.

vmezzela:
@fgyanz done!

fgyanz:
LGTM! Thanks a lot, @vmezzela and @marcosps

fgyanz (approved):

PR #259, opened 2026-08-06
utils: Fix codestream filtering

fgyanz:
filter_codestreams() incorrectly matches codestreams which share the same string start as the filter regex. For example:
	`--filter "16.0rtu1"`
returns:
	`"16.0rtu1, 16.0rtu10, 16.0rtu11, 16.0rtu12, 16.0rtu13,
	16.0rtu13, 16.0rtu15"`

Doing a full match instead fixes the issue.


vmezzela (approved):
lgtm!

marcosps:
lgtm!

PR #260, opened 2026-08-06
extract: Support 'STATIC' klpe declarations

fgyanz:
klp-ccp preserves the 'STATIC' literal token rather than expanding it, so the old regex was incorrectly stripping those declarations as "non-static."


marcosps:
Just tested, and it worked as expected. LGTM

PR #261, opened 2026-08-17
plugins/extract: Remove leftover typeofs() on cleanup

fgyanz:
`klp-cc` has a tendency to include typeofs declarations for klpe and klpp functions. Such declarations always cause conflicts during livepatch compilation.
It's safer to just remove them altogether.

Note that `extern typeofs` are not targeted by this change.


marcosps (approved):
My main concern would be the IBT klp relocs, but your regex does not remove it, so LGMT!

PR #262, opened 2026-08-17
extract: Error out when klp-ccp generated patched_funcs is empty

marcosps:


klpbuild/plugins/extract.py
```diff
@@ -320,6 +320,9 @@ def get_klpp_symbols(out_dir, lp_out):
     if not fpath.exists():
         raise RuntimeError(f"File not found: {fpath}")
 
+    if fpath.stat().st_size == 0:
+        raise RuntimeError(f"File {fpath} is empty.")
```
fgyanz: Could you perhaps also add why is it empty? This can only happen if klp-ccp couldn't find any symbol. So something like `klp-ccp could not find any of the indicated symbols in {obj}`. What to you think?

fgyanz (approved):
LGTM! thanks :)

fgyanz (approved):

PR #263, opened 2026-08-18
plugins:extract: Remove klpp symbols declarations

fgyanz:
Remove all the klpp function declarations in the C files. Otherwise there might be signature conflicts during compilation, as klp-build already declares them in the header file.

A partial solution already existed in the past, but was removed in 2ccb6e0 ("extract: Skip klpp function declarations"). This commit adds a more robust fix, which effectively removes the any duped declaration. More importantly, it does the change only after the *.orig file has been created; which didn't happen in the past.

### DIFF
--- a/klpbuild/plugins/extract.py
+++ b/klpbuild/plugins/extract.py
@@ -315,6 +315,26 @@ def get_ext_symbols(out_dir):
     return symbols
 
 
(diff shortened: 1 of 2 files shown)


marcosps (approved):
LGTM! And thanks for fixing the tests!

vmezzela (approved):
lgtm!

PR #264, opened 2026-08-19
utils: Improve fast filtering

fgyanz:
The current fast implementation filters out too many codestreams. `--fast` should push the smallest possible subset of codestreams required to verify the correctness of the livepatch. Which means filtering at least one codestream per product in each group/branch.


vmezzela (approved):
lgtm!

marcosps:
LGTM!

PR #268, opened 2026-08-27
klplib/codestream: do not override module_name when gating config is builtin

vmezzela:
When running klp-build setup with a built-in gating configuration option (such as --conf CONFIG_X86 with --module kvm), get_file_mod() would previously observe CONFIG_X86=y and override module_name to vmlinux, discarding the designated module.

Update get_file_mod() to resolve to vmlinux only if fdat.module_name is explicitly vmlinux or if the gating config is a tristate module config (is_module_on_any() is True) that happens to be built-in on the target architecture. This ensures explicit or non-module gating configs do not clobber the specified module_name.


fgyanz (approved):
Tested and review. LGTM!

PR #270, opened 2026-09-01
Add --commit option to setup and scan plugins

vmezzela:


vmezzela:
Example usage: `klp-build setup -n bsc1277416 --cve 2026-74287 --commit $(cve2sha CVE-2026-74287)`

klpbuild/klplib/ksrc.py
```diff
@@ -190,23 +190,39 @@ def store_patch(pfile, patch, savedir, savedir_idx, bc):
...
+
+    git_cmd = ["/usr/bin/git", "-C", kern_src, "grep", "-l"]
+    if cve:
+        cve_str = cve if cve.startswith("CVE-") else f"CVE-{cve}"
+        git_cmd.extend(["-e", cve_str])
+    for commit in commits:
+        git_cmd.extend(["-e", f"Git-commit: {commit}"])
+
+    if not cve and not commits:
+        return []
```
marcosps: This could be moved to the beginning of the function

klpbuild/klplib/ksrc.py
```diff
@@ -190,23 +190,39 @@ def store_patch(pfile, patch, savedir, savedir_idx, bc):
         f.write(pfile)
 
 
-def get_branch_patches(cve, mbranch):
+def get_branch_patches(cve=None, mbranch=None, commits=None):
     kern_src = get_user_path('kernel_src_dir')
 
+    if commits is None:
+        commits = []
```
marcosps, line 200: Given that --commits would always be default to [] on setup and scan, the None check can be removed, all arguments would always be passed in this case, were they not?

marcosps:
I liked the change, as it will make things easier for us. Just a couple of comments, but the code itself it good, thanks a lot!

klpbuild/klplib/ksrc.py
vmezzela: Yes, you are correct. Nonetheless keeping the check there make the code slightly more robust in case of future changes.

klpbuild/klplib/ksrc.py
vmezzela: Right, I am pushing an update with this change.

klpbuild/klplib/ksrc.py
marcosps: But mbranch here cannot be None, right? I mean, we should make sure that's not the case.

marcosps: It's nothing series, just wanted to check if I got it correctly.

vmezzela: > But mbranch here cannot be None, right? I mean, we should make sure that's not the case.

Right, `mbranch=None` shouldn't be allowed and there is one issue. I am updating it.

klpbuild/klplib/ksrc.py
```diff
@@ -190,23 +190,39 @@ def store_patch(pfile, patch, savedir, savedir_idx, bc):
...
+        return []
+
+    if commits is None:
+        commits = []
+    if isinstance(commits, str):
+        commits = [commits]
+
+    git_cmd = ["/usr/bin/git", "-C", kern_src, "grep", "-l"]
+    if cve:
+        cve_str = cve if cve.startswith("CVE-") else f"CVE-{cve}"
```
fgyanz: In which case does an input `cve` start with `CVE-` ? AFAIK we normalize all input CVE values in `utils.is_cve_valid()` . Both bugzilla and `--cve` must follow a strict format, and `CVE-` prefix is not allowed. Do you mind clarifying if there is something I'm missing?

klpbuild/klplib/ksrc.py
```diff
@@ -221,9 +237,11 @@ def get_branch_patches(cve, mbranch):
 
 
 @__check_kernel_source_tags_are_fetched
-def get_patches(cve, savedir=None, extra_patches=None):
+def get_patches(cve=None, savedir=None, extra_patches=None, commits=None):
     if extra_patches is None:
         extra_patches = []
+    if commits is None:
+        commits = []
```
fgyanz, line 243: Would it make sense to have `commits` feature enabled by default? That is, if `--commit` is not set, run `cve2sha` from klp-build and use the returned shas. Otherwise, if `--commit` is set, use the input shas instead. By doing so we can cover cases where CVE is not assigned yet, for instance.

fgyanz:
Thanks a lot! I added some minor comments, but the PR is sound. BTW don't forget to update the man page too :P 

As of now the PR is being used by the internal CI.

klpbuild/klplib/ksrc.py
vmezzela: You are right. I forgot we only pass the CVE without the prefix!

klpbuild/klplib/ksrc.py
vmezzela: I don't know, I prefer to keep `cve2sha` separate from klp-build (or at least for this PR)

fgyanz (approved):
LGTM!

klpbuild/plugins/setup.py
```diff
@@ -181,17 +193,18 @@ def setup_codestreams(lp_name, data):
...
-        if not (cve := data["cve"]):
+        if not (cve := data.get("cve")) and not data.get("commits"):
             cve = bugzilla.get_bug_cve(bugzilla.get_bug(lp_name))
             assert cve, f"Could not retrieve CVE from bugzilla for {lp_name}"
             logging.info("CVE retrieved from bugzilla: %s", cve)
             data["cve"] = cve
 
-        _, upstream, patched_cs, codestreams = scan(data["cve"], data["conf"],
+        _, upstream, patched_cs, codestreams = scan(data.get("cve"), data["conf"],
                                                     data["lp_filter"], True,
```
marcosps, line 203: data.get("cve") without a default value would fail the same way as data["cve"]?

marcosps (approved):
Besides my comment, everything looks good to me!

klpbuild/plugins/setup.py
vmezzela: No, it wouldn't fail actually. Good catch, maybe it's better to leave it as it was before. What do you think?

marcosps: I mean, at this point we would have a CVE either way. Sure, leaving it as it was before sounds good

vmezzela: fix pushed

PR #271, opened 2026-09-07
 bugzilla: Don't drop 'not affected' bugs 

fgyanz:
Bugs marked as 'not affected' by KSS shouldn't be dropped.
Those bugs will be closed, without any backport been done, because
the fixes already exists in the product branches. However, that
doesn't mean all supported codestreams include the fix.

Hence, the correct action is to mark those bugs as fixed.


vmezzela (approved):
lgtm!

marcosps:
LGTM!

